PR #2, opened 2020-03-21
Adding drills in the database

guillaumemoby:
I've added somes drills
I am trying to work an a webview


tonnenpinguin:
Hey!
Any chance you could run `npm run prettier-fix` on the changes you made?
Which Editor are you using? VSCode offers an option to automatically run prettier on every save using the `.prettierrc` configuration.

sallesma:
Maybe we could use something like https://github.com/typicode/husky to have prettier being run on any committed file without everyone needing to configure anything on their side ?

tonnenpinguin:
@sallesma Sounds pretty neat!
Do you want to take over #3 ?

PR #6, opened 2020-03-21
Make two different lists for technical and fitness drills

Adeloche:
- make two different buttons on Homepage: one to display technical drills, the other to display fitness drills

NB: description from Moby is:
" The home page: For the V0, we aim to have 3 buttons. (Technical drills, Fitness drills and Training)"
--> I don't know what is 'traing' content, so I haven't done it for now

### DIFF
--- a/src/Components/DrillListPage.js
+++ b/src/Components/DrillListPage.js
@@ -9,7 +9,9 @@ const mapStateToProps = state => {
 };
 
 const DrillListPage = props => {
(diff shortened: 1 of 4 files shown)


guillaumemoby:
The training button is to prepare a full practice. In this area, we will prepare a full practice. With warm up drill; technical and game .
The long term goal will be to help the coachs to create their own practice.

PR #17, opened 2020-03-22
Remove animation from steps + add training

Adeloche:


src/Fixtures/index.js
```diff
@@ -68,27 +62,28 @@ export default {
       duration: 10, // minutes
       goals: ['Catch', 'Handling'],
       level: ['intermediate'],
+      video: 'https://www.youtube.com/embed/oN1bzPCKkGE',
       steps: [
         {
+          key: '1',
```
sallesma: LGTM except that we can remove those keys now

Issue #19, opened 2020-03-22
Training page


Adeloche:
From Homepage, when user choose 'Training' he is redirected to a page showing all available trainings.
When he select a training, he is redirected to a page showing all drills from training

![image](https://user-images.githubusercontent.com/62174449/77257442-a0eb4200-6c74-11ea-89ff-188adf67622c.png)

sallesma:
- Add a description for the session.

Maybe #53 should be done first ?

PR #26, opened 2020-03-22
Training  page

Adeloche:
First simple version of training  page as in #19 .

**Update 04/04/2020**
Need some rework:
- traaining list page will be handled by #55 
- this PR will handle training page only
- after discussion on 04/04, we also need to display the training session description at the top


src/Components/DrillListPage.js
```diff
@@ -14,10 +14,22 @@ const secondsToMinutes = seconds => {
   return date.getMinutes();
 };
 
+const filtered_drills = data => {
+  const drills = data.drills;
+  const filter = data.filter;
+  if ('type' in filter) {
+    return drills.filter(drill => drill.type === filter.type);
+  }
+  if ('training' in filter) {
```
sallesma: I think that i you extract a `TrainingListPage` component, this block becomes useless

src/Fixtures/index.js
```diff
@@ -279,12 +279,12 @@ export default {
       steps: [],
     },
   ],
-  training: [
+  trainings: [
     {
```
sallesma: I guess a training could contain the ids of drills that are included. That would be closer to what we would do with a relational database under the hood ?

Adeloche: That is what I tried with the 'drills' field of these training, is that what you mean or something else ?

src/Components/DrillListPage.js
Adeloche: I do have a TrainingListPage component to display training objects.
But then when I select one of these training from the list, I want to display the drills of the training on something that looks like a DrillListPage, except the filtering is different (I want to display only drills in the training). That's why I think I needed this block.
Should I do a dedicated trainingPage instead ?

src/Fixtures/index.js
sallesma: I read too fast, I meant exactly what you did 😄

src/Components/DrillListPage.js
sallesma: Then maybe we should extract a `DrillList` component that is responsible for displaying a list of drills passed as props?

And leave it to `DrillListPage` and `TrainingPage` to filter which drills are to be displayed and use this `DrillList` ?

Adeloche:
Need some rework:
- training list page will be handled by #55 
- this PR will handle training page only
- after discussion on 04/04, we also need to display the training session description at the top

Adeloche:
I reworked Training page, adding only the description after our talk yesterday so it now looks like this:
image

@guillaumemoby @sallesma I would like to have you opinion on other information which could be useful to display on this page for a training session:
- should we display sessions goals as it done on drillpage ?
- what about equipement, do we display a synthesis here as well or only on drill page ? I pretty much have the same questions for players, duration and level
- what do you think of adding a START button which goes directly to the first drill ?

src/Components/DrillList.js
```diff
@@ -0,0 +1,57 @@
...
+  const navigation = props.navigation;
+  console.log(navigation);
+
+  return (
+    <View>
+      <FlatList
+        data={drillsToDisplay}
+        keyExtractor={item => item.id.toString()}
+        renderItem={({ item }) => (
+          <TouchableOpacity style={styles.drill} onPress={() => navigation.navigate('DrillPage', { drill: item })}>
```
arribbar, line 17: You should use destructuration for `styles` and `item`

arribbar: And `renderItem`, it might be interesting to write a proper function for this one

src/Components/TrainingPage.js
```diff
@@ -0,0 +1,43 @@
...
+
+class TrainingPage extends Component {
+  render() {
+    const { navigation } = this.props;
+    const training = this.props.route.params.training;
+    const drills = this.props.drills.filter(drill => training.drills.includes(drill.id));
+
+    return (
+      <ScrollView style={styles.trainingPage}>
+        <Text style={styles.descriptionText}>{training.description}</Text>
```
arribbar, line 21: You can use destructuration for `styles`

src/Components/DrillList.js
Adeloche: done

src/Components/TrainingPage.js
```diff
@@ -1,21 +1,43 @@
-import React from 'react';
-import { StyleSheet, View, Text } from 'react-native';
+import React, { Component } from 'react';
+import { StyleSheet, ScrollView, Text } from 'react-native';
 import { connect } from 'react-redux';
+import DrillList from './DrillList';
 import theme from '../styles/theme.style';
 
+const getGoals = trainingDrills => {
+  const nbPlayersList = trainingDrills.map(({ nbPlayers }) => nbPlayers);
```
sallesma, line 8: I guess we should bring the goals instead of nbPlayers?

src/Components/TrainingPage.js
```diff
@@ -1,21 +1,43 @@
...
-};
+class TrainingPage extends Component {
+  render() {
+    const { navigation } = this.props;
+    const training = this.props.route.params.training;
+    const drills = this.props.drills.filter(drill => training.drills.includes(drill.id));
+
+    return (
+      <ScrollView style={styles.trainingPage}>
+        <Text style={styles.descriptionText}>{training.description}</Text>
```
sallesma, line 31: Maybe add a training title?

src/Components/TrainingPage.js
```diff
@@ -1,21 +1,43 @@
...
+import DrillList from './DrillList';
 import theme from '../styles/theme.style';
 
+const getGoals = trainingDrills => {
+  const nbPlayersList = trainingDrills.map(({ nbPlayers }) => nbPlayers);
+  return Math.max(...nbPlayersList);
+};
+
+const getEquipment = trainingDrills => {
+  const nbPlayersList = trainingDrills.map(({ nbPlayers }) => nbPlayers);
```
sallesma, line 13: Same with equipment?

Adeloche:
> I guess withif you merge master in this branch and make a few corrections we could merge this PR to have a first version of this page, and then iterate later?

I haven't finished adding the training basic information  we discussed last time (goals, equipement, proper description) yet. I'll try working on it this week to have a first version, I'll request a new review then.

src/Components/TrainingPage.js
Adeloche: For now the training title is displayed in the header bar, as for drills, so I haven't put it again on the page. I think I will make another PR to improve the page design which is ugly for now.. I would be very happy to have some suggestions

src/Components/TrainingPage.js
Adeloche: I'm not sure how we want to use and store equipment for now..so I leave it for now. I would like to discuss this so I can add it in another PR

Adeloche:
> I guess withif you merge master in this branch and make a few corrections we could merge this PR to have a first version of this page, and then iterate later?

Ok for me, I listed the improvements to have what we discussed for V0  in #73.

src/Components/TrainingPage.js
sallesma: Yes that's right
I guess we'll have better ideas when the page is in the app for some time and we navigate it from time to time

src/Components/DrillList.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
```
sallesma, line 1: Maybe we could move this file to the `shared` subfolder ?

Adeloche: sound good, done

PR #31, opened 2020-03-24
Technical drills filter

sallesma:
I'd be happy to get reviews on this one


src/Components/DrillListPage.js
```diff
@@ -13,27 +14,39 @@ const DrillListPage = props => {
...
-          </TouchableOpacity>
-        )}
-        onEndReachedThreshold={0.5}
-        onEndReached={() => {}}
-      />
+      <Text style={styles.counter}>{data.length} drills available</Text>
+      {displayFilters ? (
+        <Filters
+          onConfirm={() => setDisplayFilters(false)}
+          onFiltered={drills => setData(drills)}
```
sallesma: My idea here is that the responsibility for filtering should not be on the List component but on the Filters (they know what are the filters)

Also this mean we can extend the code adding filters with no impact on this component?

But I'm not sure we can test the behaviour of one without the other one?

Any thoughts on this?

PR #38, opened 2020-03-28
Technical drills filter

### LINKED ISSUE #16: Drill List - Filtering technical drills
When the user taps the filter button on the Drill List, it opens a filters view

Fitler button could look like :
Screenshot 2020-03-22 at 19 19 54

Filters view :
Screenshot 2020-03-22 at 19 18 59

- The back arrow should take us back to the exercise list and not the homepage.
- Add back arrows and validation as in Nike Training Club (see examples [here](https://github.com/UltimateApp/UltimateApp/wiki))

sallesma:
Closes #16


package.json
```diff
@@ -5,7 +5,6 @@
     "android": "expo start --android",
     "ios": "expo start --ios",
     "web": "expo start --web",
-    "eject": "expo eject",
```
sallesma: It seems to me that we don't need this

src/Components/Filters.js
```diff
@@ -11,25 +11,52 @@ const Button = props => {
...
-  const [level, setLevel] = useState();
+  const [selectedLevel, setLevel] = useState();
+  const [selectedGoal, setGoal] = useState();
 
   const onLevelChange = pressedLevel => {
-    const newLevel = pressedLevel === level ? undefined : pressedLevel;
+    const newLevel = pressedLevel === selectedLevel ? undefined : pressedLevel;
     setLevel(newLevel);
   };
+  const onGoalChange = pressedGoal => {
```
sallesma: I just duplicated code to create the second filter
I feel like we should refactor this when implementing the third one (rule of three)

sallesma: Ok I now have 3 filters working, with some nice duplicated code to refactor

src/Components/Filters.test.js
```diff
@@ -5,7 +5,40 @@ import Filters from './Filters';
 
 describe('<Filters />', () => {
   it('renders correctly', () => {
-    const tree = renderer.create(<Filters initialData={[]} onConfirm={jest.fn()} onFiltered={jest.fn()} />).toJSON();
+    const drills = [
```
tonnenpinguin, line 8: I would move testdata into a separate file called something along the lines of  `testFactory`. Then we can reuse the same datastructures for other tests a lot more easily.

When you setup the factory make sure to make all values configurable like so:
```
export const createDrill = ({ id = 10, type = "fitness", [...]}) => ({
    id,
    type,
    [...]
});
```

src/Components/Filters.js
tonnenpinguin: Have you ever worked with ramda (https://ramdajs.com/)?

It takes a bit to wrap your head around, but it's insanely useful for writing reusable functional (and immutable!) code (like filters).

As an example the filter code below could be rewritten to something like this:
```
const newData = R.pipe(
     R.filter(R.propEq('level', selectedLevel)),
     R.filter(R.propSatisfies('goals', R.includes(selectedGoal))),
     R.filter(R.propSatisfies('minimalPlayersNumber', R.lte(R.__, numberOfPlayers)))
)(initialData);
```

src/Components/Filters.js
```diff
@@ -60,6 +97,16 @@ const styles = StyleSheet.create({
...
   },
+  input: {
+    marginBottom: 30,
+    padding: 10,
+    paddingHorizontal: 20,
+    fontSize: 16,
+    color: '#444',
+    borderBottomWidth: 1,
+    borderColor: '#ddd',
+    backgroundColor: '#F5F5F5',
```
tonnenpinguin, line 108: We should probably start using a central file to manage colors, otherwise this is going to be one hell of a rainbow :)

src/Components/Filters.js
```diff
@@ -11,26 +11,60 @@ const Button = props => {
...
+          onPress={() => onLevelChange('intermediate')}
+          active={selectedLevel === 'intermediate'}
+        />
+        <Button title="Advanced" onPress={() => onLevelChange('advanced')} active={selectedLevel === 'advanced'} />
+      </View>
+      <Text style={styles.filterTitle}>Goals</Text>
+      <View style={styles.filter}>
+        {props.initialData
+          .map(drill => drill.goals)
+          .reduce((x, y) => x.concat(y), [])
```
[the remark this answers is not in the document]
sallesma: I used it previously but the tests would not pass. Maybe because flatMap is ES2019 ?

src/Components/Filters.test.js
sallesma: Totally agree!

src/Components/DrillListPage.test.js
```diff
@@ -39,7 +39,7 @@ describe('<DrillListPage />', () => {
         image: 'http://www.liberte-fitness.fr/sites/default/files/styles/slider/public/news/tone.jpg?itok=505bme2a',
         description:
           'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec scelerisque dignissim felis, at tristique mi efficitur a. Nulla pellentesque odio nunc, ut sodales ex pulvinar at. In hac habitasse platea dictumst. Pellentesque sodales nisl lorem, ac lacinia nisl lacinia eu. Proin lacinia viverra mauris, et pharetra ipsum tempus eget. Ut maximus sapien in hendrerit ultricies. Maecenas vel diam tincidunt, consequat nunc a, mattis eros. Curabitur a eros a nulla malesuada gravida. Donec massa quam, suscipit eget efficitur et, euismod sed nulla. Morbi mattis et magna a aliquam. Fusce pellentesque vel erat eget',
-        nbPlayers: 2,
+        minimalPlayersNumber: 2,
```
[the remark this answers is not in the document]
sallesma: `count` is more precise since it refers to something countable and not just a random number.

I think `number` makes more sense for a french developer since it's closer to the french word `nombre`

I'm ok with count

src/Components/Filters.js
sallesma: I did'nt know it. It seems nice and could be used in `useEffect` below. My concern here is I'm not very happy with defining `  const [x, setX] = useState();` and `onXChange` for each filter and thinking about refactoring it.

But any implementation I imagine is a working abstraction bt makes it more difficult to understand (and modify or extend) than the actual duplication. I think I'll stick with the duplication :)

src/Components/Filters.js
sallesma: I did this in another pull request: https://github.com/UltimateApp/UltimateApp/pull/49

src/Components/Filters.js
tonnenpinguin: Worst case there's always class components with a single `setState` function. You could probably leave the filtering in the render function as it only renders on state/prop updates anyway

src/Components/Filters.js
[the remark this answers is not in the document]
tonnenpinguin: Ah yeah, that might be a problem. I mean flatMap is available on ~86% percent of browsers (most of the outliers being surprise, surprise IE, some older Edge versions and obscure mobile browsers).
We could probably change the targeted ECMA version to something higher to get the tests to pass. On the other hand you could actually use some ramda magic for this, too:

```
R.pipe(
    R.map(R.prop("goals")),
    R.flatten
)(initialData)
```

src/Components/Filters.test.js
sallesma: I did it on moby's pull request #45 
I'll merge it here when it's merged

arribbar:
Where are we on this PR ? 
It seems there are some conflicts right now ^^

sallesma:
Yep, I'm working on it on my local branch
- better pages flow between filters and the rest of the app
- use enums
- fix conflicts

The I think we could merge this and move on to work on implementing separate filters for fitness/trainings/technical in another PR

arribbar:
Some conflict to resolve :)

arribbar:
...
Same same
... ...

src/Components/Filters.test.js
```diff
@@ -1,31 +1,53 @@
...
 
-      await fireEvent.press(getByText('Beginner'));
+      await fireEvent.press(getByText(Levels.ADVANCED));
 
-      expect(onFiltered).toBeCalledWith([]);
+      expect(getByTestId('availableDrills').props.children.join('')).toEqual('1 drills available');
 
-      await fireEvent.press(getByText('Validate'));
+      // await fireEvent.press(getByText('Validate'));
+      // TODO: Need to access the header button. Or to find another solution
```
sallesma, line 47: I need to fix this before we can merge this pull request

src/Components/TrainingListPage.js
```diff
@@ -9,9 +9,9 @@ const getTrainingDuration = trainingDrills => {
   return durationList.reduce((a, b) => a + b, 0);
 };
 
-const getTrainingNbPlayers = trainingDrills => {
-  const nbPlayersList = trainingDrills.map(({ nbPlayers }) => nbPlayers);
-  return Math.max(...nbPlayersList);
+const getTrainingminimalPlayersNumber = trainingDrills => {
```
arribbar, line 12: missing uppercase for M `getTrainingMinimalPlayersNumber`

src/Navigation/Navigation.js
```diff
@@ -1,18 +1,19 @@
...
+import Filters from '../Components/Filters';
 import TrainingListPage from '../Components/TrainingListPage';
 import TrainingPage from '../Components/TrainingPage';
 
 const Stack = createStackNavigator();
 
 export const Navigation = () => (
   <NavigationContainer>
-    <Stack.Navigator>
+    <Stack.Navigator screenOptions={{ cardStyleInterpolator: CardStyleInterpolators.forHorizontalIOS }}>
```
arribbar, line 16: What does your change ? 🤔

sallesma: Next screen comes from the left instead of the bottom, or something like this.

I'm not sure it makes a difference. That's from when I was trying some stuff to display filters differently.

src/Components/Filters.js
```diff
@@ -11,44 +12,133 @@ const Button = props => {
...
+          {this.state.displayedDrills.length} drills available
+        </Text>
+        <View style={styles.filters}>
+          <Text style={styles.filterTitle}>Level</Text>
+          <View style={styles.filter}>
+            {Object.values(Levels).map(level => (
+              <Button
+                key={level}
+                title={level}
+                onPress={this.onPressedChange.bind(this, 'selectedLevel', level)}
```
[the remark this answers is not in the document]
sallesma: It's a bit annoying because I would have to bind quite a lot of functions in the constructor

... which means the performance issues would be even more annoying if we keep redefining functions here, I guess?

src/Navigation/Navigation.js
sallesma: So yes, it makes new pages "come" from the right side of the screen instead of the bottom. What do you prefer?

(I have no idea what's best)

src/Components/Filters.js
```diff
@@ -11,44 +12,133 @@ const Button = props => {
...
+    if (selectedGoal) newData = newData.filter(drill => drill.goals.includes(selectedGoal));
+    if (numberOfPlayers) newData = newData.filter(drill => drill.minimalPlayersNumber <= numberOfPlayers);
+
+    this.setState({ displayedDrills: newData });
+  }
+
+  componentDidMount() {
+    this.props.route.params.onFiltered(this.state.displayedDrills);
+    this.props.navigation.setOptions({
+      headerRight: () => (
```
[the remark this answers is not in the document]
sallesma: The function or the component being rendered?

src/Navigation/Navigation.js
sallesma: I removed it, let's keep it simple

src/Components/Filters.js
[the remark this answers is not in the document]
sallesma: I extracted a HeaderButton

src/Components/Filters.js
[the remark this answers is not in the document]
arribbar: Yeah, the other solution is to make an arrow function.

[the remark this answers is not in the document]
sallesma: It's done
Tell me if it's ok :)

src/Components/Filters.js
```diff
@@ -1,54 +1,154 @@
...
+          }}
+        />
+      ),
+    });
+  }
+
+  render() {
+    const { selectedLevel, selectedGoal, numberOfPlayers } = this.state;
+    return (
+      <View style={styles.wrapper}>
```
arribbar, line 79: To better readability and reuse, you could split this big `render` function in some shorter one. Well, this is up to you ;)

src/Components/Filters.js
```diff
@@ -1,54 +1,154 @@
...
+                <Button
+                  title={goal}
+                  onPress={() => this.onPressedChange('selectedGoal', goal)}
+                  key={goal}
+                  active={selectedGoal === goal}
+                />
+              ))}
+          </View>
+          <Text style={styles.filterTitle}>Number of players: {numberOfPlayers}</Text>
+          <Slider
```
arribbar, line 109: I do not why but on my phone, when I use the slider. The animation does not work well when I keep my finger on it

src/Components/DrillListPage.js
```diff
@@ -1,66 +1,66 @@
...
-      {!displayFilters && (
-        <TouchableOpacity style={styles.filterButton} onPress={() => setDisplayFilters(true)}>
-          <Text style={styles.filterButtonText}>Filter</Text>
-        </TouchableOpacity>
-      )}
+      <Text style={list.counter}>{displayedDrills.length} drills available</Text>
+      <FlatList
+        data={displayedDrills}
+        keyExtractor={item => item.id.toString()}
+        renderItem={({ item }) => (
```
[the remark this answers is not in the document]
sallesma: Yep, Adélie started it on her branch, I didnt want to duplicate the refactoring

src/Components/DrillListPage.js
```diff
@@ -1,66 +1,66 @@
...
-
 export const DrillListPage = props => {
   const { navigation } = props;
   const type = props.route.params.type;
-  const drills = props.drills.filter(drill => drill.type === type);
 
-  const [data, setData] = useState(drills);
-  const [displayFilters, setDisplayFilters] = useState(false);
+  const storeDrills = props.storeDrills.filter(drill => drill.type === type);
+  const displayedDrills = props.route.params.filteredDrills ? props.route.params.filteredDrills : storeDrills;
```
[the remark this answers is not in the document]
sallesma: It's `[]`

src/Components/Filters.js
sallesma: I was thinking about refactoring the filters component. But I wanted to build the Fitness Filters before to get a wider idea of what could be the right abstraction (or even trainings filters to follow the rule of three)

src/Components/Filters.js
sallesma: Weird...

I was planning to move to another library with the fitness filters to be able to choose a range for the duration of drills. (maybe https://github.com/ptomasroos/react-native-multi-slider)

Moreover Slider is a deprecated react-native feature, but the up-to-date library is not yet available in expo apps. I think we can keep a "not perfect" slider for now and move quickly hen a better on is available. More info there : https://github.com/react-native-community/react-native-slider/issues/152#issuecomment-593786774

PR #39, opened 2020-03-28
Favorite

guillaumemoby:
I am an issue with the favorite button. I feel I have to initiate the state in DrillPage.js with something looking like :
  constructor(props) {
    super(props);
    this.state = { drill: props.drill };
  }
but it doesn't work at the moment. The error change but I am not sure that I am going to the direction. Any help


src/Components/DrillPage.js
```diff
@@ -1,25 +1,32 @@
 import React, { Component } from 'react';
-import { ScrollView, StyleSheet, View, Text, Button, ImageBackground, TouchableOpacity } from 'react-native';
+import { ScrollView, StyleSheet, View, Text, Image, ImageBackground, TouchableOpacity } from 'react-native';
 import { connect } from 'react-redux';
 
 class DrillPage extends React.Component {
+  _toggleFavorite() {
+    console.log('TOGGLE FAVORITE function ', this.state.drill);
```
sallesma: Here you cannot use `this.state.drill` because it had not been initialized (as you said in your message)

But you don't need to put the drill in the state juste to be able to call it here. We use the drills in the props to render the view. Maybe fetch it the same way ?
`this.props.route.params.drill`

src/Components/DrillPage.js
```diff
@@ -1,25 +1,32 @@
...
+    console.log(this.props.favoritesDrill);
+  }
 
-componentDidUpdate() {
-  console.log("componentDidUpdate : ")
-  console.log(this.props.favoritesDrill)
-}
+  _displayFavoriteImage() {
+    var sourceImage = require('../Images/ic_favorite_border.png');
+    if (this.props.favoritesDrill.findIndex(item => item.id === this.state.drill.id) !== -1) {
```
sallesma: Same here, maybe use the drill present in the props
` this.props.route.params.drill`

src/Components/DrillPage.js
```diff
@@ -1,25 +1,32 @@
...
+      // Film dans nos favoris
+      sourceImage = require('../Images/ic_favorite.png');
```
sallesma: I find it cleaner to import the file at the beginning of the file as done in HomePage.js :

`import fitness from '../../assets/HomePage/fitness.png';`

src/Components/DrillPage.js
```diff
@@ -1,25 +1,32 @@
 import React, { Component } from 'react';
-import { ScrollView, StyleSheet, View, Text, Button, ImageBackground, TouchableOpacity } from 'react-native';
+import { ScrollView, StyleSheet, View, Text, Image, ImageBackground, TouchableOpacity } from 'react-native';
 import { connect } from 'react-redux';
 
 class DrillPage extends React.Component {
+  _toggleFavorite() {
```
sallesma: By the way it looks like this function is not used, you may remove it and use only `this.props.toggleFavorite(drill)`

src/Components/DrillPage.js
```diff
@@ -1,25 +1,24 @@
...
-}
-
-componentDidUpdate() {
-  console.log("componentDidUpdate : ")
-  console.log(this.props.favoritesDrill)
-}
+  constructor(props) {
+    super(props);
+    this.state = { drill: props.drill };
+  }
```
tonnenpinguin: I prefer the more modern variant of intitializing the state:

```
state = {
  drill: this.props.drill
}
```

Use this instead of the constructor

src/Components/DrillPage.js
```diff
@@ -33,32 +32,32 @@ componentDidUpdate() {
...
           <TouchableOpacity
             style={styles.videoLink}
-            onPress={() => navigation.navigate('DrillAnimationPage', { drill: drill })}
+            onPress={() => navigation.navigate('DrillAnimationPage', { drill })}
           >
             <Text style={styles.videoLinkText}>Video</Text>
           </TouchableOpacity>
         </ImageBackground>
         <View style={styles.separator} />
+        <TouchableOpacity style={styles.favoriteContainer} onPress={() => this.props.toggleFavorite(drill)}>
```
tonnenpinguin: Destructure all the props at once and use the local variable afterwards, this makes it a lot easier to figure out which props are used in a component

src/Fixtures/index.js
```diff
@@ -128,7 +964,7 @@ export default {
       equipment: 'One disc, 2 cones, 3 players',
       durationInMinutes: 18,
       goals: ['Offense', 'Handling'],
-      level: 'beginner',
+      level: 'Beginner',
```
tonnenpinguin: I would suggest using enums for all variables that have a fixed set of values. It prevents difference in Spelling, makes it a lot easier to figure out where something is used and makes auto completion work a lot better

https://www.sohamkamani.com/blog/2017/08/21/enums-in-javascript/

src/Components/DrillAnimationPage.js
```diff
@@ -12,7 +12,7 @@ class DrillPage extends React.Component {
         {drill.animation ? (
           <Animation animation={drill.animation} />
         ) : drill.video ? (
-          <WebView source={{ uri: 'https://www.youtube.com/embed/oN1bzPCKkGE' }} style={{ marginTop: 20 }} />
+          <WebView source={{ uri: drill.video }} style={{ marginTop: 20 }} />
```
arribbar: I would set style in a constant as done [line 24](url)

src/Components/DrillListPage.js
```diff
@@ -29,10 +30,15 @@ const DrillListPage = props => {
       ) : (
         <FlatList
           data={data}
+          /* extraData={this.props.actions.favoritesDrill} */
```
arribbar: Delete the comment if you do not need it

src/Components/DrillPage.js
arribbar: For @guillaumemoby You can destructure (since es5) like this:
```
const {
  favoriteContainer, 
  videoLinkText,
} = styles;
```
Then, you can access `favoriteContainer` as a const in your code

src/Fixtures/index.js
```diff
@@ -276,10 +1112,10 @@ export default {
       ],
     },
     {
-      id: 13,
+      id: 51,
       type: 'fitness', // vs technical
-      source: 'Rise Up',
-      title: 'Hot Box',
+      source: '',
+      title: 'Yoga etirement',
```
arribbar: Yoga Stretching, no ? ;)

src/Store/Reducers/favoriteReducer.js
```diff
@@ -1,29 +1,32 @@
 // Store/Reducers/favoriteReducer.js
 
-const initialState = { favoritesDrill: [] }
+const initialState = { favoritesDrill: [] };
 
 function toggleFavorite(state = initialState, action) {
-  let nextState
+  let nextState;
+
+  console.log(state);
```
arribbar: Once you finish your test, you can delete these two logs (or set a better to comment to understand them)
But I think you can delete them

arribbar:
Oh, and be careful, you have two conflicts with master ... You need to resolve them

src/Fixtures/index.js
guillaumemoby: yes

guillaumemoby: For the moment the Index.js doesn't have a good structure. I will complete with real data later

src/Fixtures/index.js
sallesma: I think introducing enums is an awesome idea
But we should do this in another pull request since it requires a bit of refactoring

src/Components/DrillPage.js
guillaumemoby: @arribbar : I could need deeper explanation for this point

tonnenpinguin:
@arribbar What do you think about the following convention:

* props and state should be destructured at the beginning of a function
* styles should not be destructured to make it clear that we're referring to a style (as opposed to a data object/function)

src/Components/DrillAnimationPage.js
```diff
@@ -25,6 +25,9 @@ const styles = StyleSheet.create({
   DrillAnimationPage: {
     height: '100%',
   },
+  VideoPresentation: {
```
tonnenpinguin, line 28: IMHO style props should be camelCase, as PascalCase is usually used for Components, Classes and Enums

src/Components/DrillAnimationPage.js
```diff
@@ -25,6 +25,9 @@ const styles = StyleSheet.create({
   DrillAnimationPage: {
```
arribbar, line 25: Can you change this var name to be in camelCase as well, please ?

arribbar: And you can use destructuration as well for `styles`

src/Components/DrillPage.js
```diff
@@ -1,22 +1,41 @@
...
-
-  componentDidUpdate() {
-    console.log('componentDidUpdate : ');
-    console.log(this.props.favoritesDrill);
+class DrillPage extends React.Component {
+  state = {
+    drill: this.props.drill,
+  };
+  _displayFavoriteImage() {
+    var sourceImage = ic_favorite_border;
```
arribbar, line 13: Instead of `var`, use `let` or `const`. Here, `let` is the correct option

src/Components/DrillPage.js
```diff
@@ -1,22 +1,41 @@
...
+    return <Image style={styles.favoriteImage} source={sourceImage} />;
   }
 
   render() {
     const drill = this.props.route.params.drill;
     const { navigation } = this.props;
+    const {
+      title,
+      infoWrapper,
+      DrillPage,
```
arribbar, line 26: You need to use the correct convention, in camelCase, so `drillPage` instead of `DrillPage`

src/Components/DrillPage.js
arribbar: Then, you call the correct style by doing simply a `<TouchableOpacity style={favoriteContainer} ... />`

src/Components/DrillAnimationPage.js
sallesma: I'm not sure about styles destructuration

Discussing it there: https://github.com/UltimateApp/UltimateApp/pull/54#discussion_r403677912

src/Components/DrillPage.js
```diff
@@ -1,62 +1,79 @@
...
-      <ScrollView style={styles.DrillPage}>
-        <ImageBackground source={{ uri: drill.image }} style={styles.image} imageStyle={styles.imageOpacity}>
-          <Text style={styles.title}>{drill.title}</Text>
-          <View style={styles.infoWrapper}>
-            <Text style={styles.info}>{drill.durationInMinutes} minutes</Text>
-            <View style={styles.separator} />
-            <Text style={styles.info}>{drill.nbPlayers} players</Text>
-            <View style={styles.separator} />
-            <Text style={styles.info}>{drill.level} level</Text>
+      <ScrollView style={DrillPage}>
```
sallesma, line 41: > I'm not sure about styles destructuration
> 
> Discussing it there: [#54 (comment)](https://github.com/UltimateApp/UltimateApp/pull/54#discussion_r403677912)

This is exactly the kind of problems we are going to run into if we destructure styles. We can't see the difference anymore between objects useful for the components behaviour (such as extracted from props, state, imported from other modules) from style objects

I feel like we have thrown away clarity and it makes the component much more difficult to read.
Would be nice to have more advices on the topic before merging this

src/Components/DrillAnimationPage.js
sallesma: But the original comment about camelCase still stands :)

src/Components/DrillListPage.js
```diff
@@ -8,6 +8,7 @@ import { DrillTypes } from '../Fixtures';
 const mapStateToProps = state => {
   return {
     drills: state.drills,
+    favoritesDrill: state.favoritesDrill,
```
sallesma, line 11: Looks like we don't use it?

src/Components/DrillPage.js
```diff
@@ -1,25 +1,28 @@
...
-  }
-
-  componentDidUpdate() {
-    console.log('componentDidUpdate : ');
-    console.log(this.props.favoritesDrill);
+class DrillPage extends React.Component {
+  state = {
+    drill: this.props.drill,
+  };
+  _displayFavoriteImage() {
```
sallesma, line 12: A new line before a class method would be nice

src/Components/DrillPage.js
arribbar: Still this comment ... ;)

src/Components/DrillPage.js
```diff
@@ -32,25 +42,30 @@ export class DrillPage extends Component {
           </TouchableOpacity>
         </ImageBackground>
         <View style={styles.separator} />
-
-        <Button title="Favoris" onPress={() => this._toggleFavorite()} />
+        <TouchableOpacity
+          style={styles.favoriteContainer}
+          onPress={() => this.props.toggleFavorite(drill)}
```
arribbar, line 47: You can probably directly call `this.props.toggleFavorite` without the arrow function

arribbar:
> @arribbar What do you think about the following convention:
> 
> * props and state should be destructured at the beginning of a function
> * styles should not be destructured to make it clear that we're referring to a style (as opposed to a data object/function)

Yep, I agree with that. We'll see at the end of v0 to refactor everything in this way

src/Components/DrillPage.js
guillaumemoby: Done for var
The second one doesn't work

guillaumemoby:
@arribbar , @tonnenpinguin and @sallesma : I've worked on your change request with Bart. I hope everything is good now

PR #45, opened 2020-03-31
Enums

guillaumemoby:
As @tonnenpinguin suggested, I have added the enums.
I don't know if we should split the drills between the frisbee and fitness.

Maybe this difference should more appears on the drillpage.js and the Drilllistpage.js


src/Fixtures/index.js
```diff
@@ -1,27 +1,66 @@
...
-  /*   const types: {
-    FITNESS: 'fitness',
-    TECHNICAL: 'technical',
-  }, */
+export const types = {
+  FITNESS: 'fitness',
+  TECHNICAL: 'technical',
+};
+
+export const intensity = {
```
sallesma: I think enum names should be plural, like you did for types

guillaumemoby: You are right. Thanks.
I want to upload only the file index.js, is there a easy way to only add this one?

arribbar:
@guillaumemoby 
You have three things to do before we can merge this PR:
- Update your branch with master. When you create a new branch, always create it when you are on master and this one is up to date
- Resolve conflicts. You can see when you do your PR that you have conflicts with master. It means we cannot merge it. To resolve conflicts, from enums branch, you need to do a `git pull origin master` to get last commits from master. Then, resolve conflicts, commit these changes and update your branch on Github
- Isolate your last commit from the previous ones

I can help you or you ask to @sallesma

tonnenpinguin:
The easiest way might be to create a new branch from master and cherry pick the commit with your changes. Then submit the new branch and open a new PR.

src/Fixtures/index.js
```diff
@@ -1,11 +1,892 @@
...
+
+export const seasonTimings = {
+  OFFSEASON: 'Off-season',
+  PRESEASON: 'Pre-season',
+  INSEASON: 'In-season',
+  ANYTIME: 'Anytime',
+};
+
+export const equipmentLabels = {
+  NOEQUIPMENT: 'No Equipment',
```
tonnenpinguin, line 43: use underscores for better readability

src/Fixtures/index.js
```diff
@@ -1,11 +1,892 @@
...
+};
+
+export const seasonTimings = {
+  OFFSEASON: 'Off-season',
+  PRESEASON: 'Pre-season',
+  INSEASON: 'In-season',
+  ANYTIME: 'Anytime',
+};
+
+export const equipmentLabels = {
```
tonnenpinguin, line 42: I would actually upcase the names to make it clearer when the code code is called (--> `EquipmentLabels`)

src/Fixtures/index.js
```diff
@@ -1,11 +1,892 @@
 import animationMenageATrois from './AnimationMenageATrois';
 import animationSquare from './AnimationSquare';
 
+export const types = {
```
tonnenpinguin, line 4: types is a bit too generic.

src/Fixtures/index.js
```diff
@@ -1,11 +1,892 @@
...
+};
+
+export const equipmentLabels = {
+  NOEQUIPMENT: 'No Equipment',
+  BASIC: 'Basic Equipment',
+  FULLEQUIPMENT: 'Full equipment',
+};
+
 export default {
+  favoritesDrill: [],
```
tonnenpinguin, line 49: `drillFavorites` or `favoriteDrills` might be a better name

src/Fixtures/index.js
```diff
@@ -45,13 +926,13 @@ export default {
       equipment: 'One disc, 2 cones, 3 players',
       durationInMinutes: 10,
       goals: ['Defense', 'Handling'],
-      level: 'beginner',
+      level: 'Beginner',
```
tonnenpinguin, line 929: use enum

src/Fixtures/index.js
sallesma: Maybe `categories` ?

src/Fixtures/index.js
sallesma: `favoriteDrills` looks better to me

guillaumemoby:
I ve apply the change that you have suggest but I can't commit anymore because of the test.

src/Fixtures/index.js
sallesma: but we should remove it here, it should be added in the other pull request

guillaumemoby:
@Martin est ce que je dois encore enlever " favoritesDrill: []," sachant que dans master il est aussi présent ?

PR #47, opened 2020-03-31
Add tests

sallesma:

I'd be happy to get feedback on this before going further.
I aim to provide behaviour tests for our components (I think snapshots are not enough), but I'm not sure wether we should go for integration tests directly ?


tonnenpinguin:
What do you guys think about instead of writing all these snapshot tests manually we could start using storybook (https://storybook.js.org/) in combination with storyshots (https://storybook.js.org/docs/testing/structural-testing/#using-storyshots).

That way we both get the benefit from storyshot (develop components in isolation or grouped, better testability because of the encouraged capsulation, better documentation) as well as automated snapshot testing with only a one-time setup being necessary

arribbar:
Well, I am not an expert to all these testing techno and will be very happy to learn these new techno.
We just need to be careful to explain well how it works to beginners...
What do you think ?@sallesma 
@tonnenpinguin If it is ok for @sallesma, do you think you can do all the setup and write some lines about how to use it ?

tonnenpinguin:
Sure thing, I can try to find some time over the weekend!

sallesma:
Seems nice to me.
I wanted to try storybook at work so I'd be glad to start here
I leave the pull request opened then

src/Components/DrillListPage.test.js
```diff
@@ -1,19 +1,51 @@
...
-    const route = {
-      params: {
-        type: 'technical',
+  const route = {
+    params: {
+      type: 'technical',
+    },
+  };
+  const drill = {
+    id: 10,
```
sallesma, line 18: Just waiting for the PR on enums #45 to DRY it

arribbar:
Some conflict to resolve :)

sallesma:
Yep, i just fixed them. Since everybody is active these days and the project is quite small we're all changing the same files :)

sallesma:
Opened an issue to keep track of the storybook idea : #56

PR #50, opened 2020-04-03
[WIP] Introduce graphql server

aenario:
Not yet used by the application itself


server/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "server",
+  "version": "1.0.0",
+  "description": "Server",
+  "main": "index.js",
+  "scripts": {
+    "test": "mocha ./test/**"
```
sallesma, line 7: Maybe we should stick to jest as a testing framework?

server/package.json
```diff
@@ -0,0 +1,17 @@
...
+  },
+  "author": "",
+  "license": "MIT",
```
sallesma, line 10: There is probably a decision to take regarding licenses. I think the expo project has none at the moment

server/README.md
```diff
@@ -0,0 +1,22 @@
...
+```
+git clone git@github.com:UltimateApp/UltimateApp.git
+cd UltimateApp/server
+npm install
+node index.js
+```
+
+## TODO
+
+- [ ] Use it in the app
```
sallesma, line 18: The big question is what for? I don't think we should use it just to use it.

What do we need this for?

server/README.md
```diff
@@ -0,0 +1,22 @@
+# 🥏 UltimateApp Server
+
+Server to host the drills for ultimate app based on Mongodb, ApolloServer and Graphql.
+
+## ⌨️ Development
+
+You need a recent version of nodejs, and a [running local mongodb server](https://docs.mongodb.com/manual/installation/#mongodb-community-edition-installation-tutorials)
+
+```
+git clone git@github.com:UltimateApp/UltimateApp.git
```
sallesma, line 10: I'm wondering whether it should be stored in another repository? What are the good practices on this?

My first inclination would be to have 1 repo for 1 app

What do you think about it ?

Adeloche: > I'm wondering whether it should be stored in another repository? What are the good practices on this?
> 
> My first inclination would be to have 1 repo for 1 app
> 
> What do you think about it ?


My opinion on that is as long as it does not become too complex, and everyone works on both parts, there are some good sides to keep it in one repo:
- everybody have an easy access to both components --> better for a good understanding of the whole thing
- no need to make 2 different PRs to introduce some functional changes
- no need of synchronisation for build/test/deployment

Moreover we can still split it later if we feel having one repo is too heavy

sallesma: wow that's a lot of good arguments in favor of keeping it here. Makes sense :)

server/package.json
aenario: Sure, there is no tests so far, I will use jest when introducing them  !

server/package.json
aenario: Ok, il will remove it

server/schema.graphql
```diff
@@ -0,0 +1,141 @@
...
+}
+
+enum Intensities {
+  LOW
+  MODERATE
+  HIGH
+}
+
+enum GoalsFrisbee {
+  DEFENSE
```
tonnenpinguin, line 47: Careful with using this kind of enums. Because each key corresponds to a numeric value things start to get very messy as soon as you either remove or add new entries. Personally I prefer to use [string enums](https://www.typescriptlang.org/docs/handbook/enums.html#string-enums) as they are a lot easier to maintain in the long run

aenario:
Closing for now, this is noise

PR #54, opened 2020-04-04
Tweak drills list

### LINKED ISSUE #51: Tweak drill List
- Add opacity on the photos
- technical drills : display number of players on the pictures (see exemple [here](https://github.com/UltimateApp/UltimateApp/wiki)) and display objectives instead of duration and number of players below drill name
- fitness drills : display drill duration on photo and display objectives instead of duration and number of players below drill name

sallesma:

 fixes #51


src/Components/DrillListPage.test.js
```diff
@@ -24,12 +24,53 @@ describe('<DrillListPage />', () => {
...
 
-  it('renders correctly with a drill', () => {
+  it('renders correctly with a technical drill', () => {
     const route = {
       params: {
         type: 'technical',
       },
     };
+    const drills = [
+      {
```
sallesma, line 34: Waiting for the enum #45 PR to be merged on master to use it here

src/Components/DrillListPage.js
```diff
@@ -33,13 +37,14 @@ export const DrillListPage = props => {
           keyExtractor={item => item.id.toString()}
           renderItem={({ item }) => (
             <TouchableOpacity style={styles.drill} onPress={() => navigation.navigate('DrillPage', { drill: item })}>
-              <Image style={styles.image} source={{ uri: item.image }} />
+              <ImageBackground source={{ uri: item.image }} style={styles.image} imageStyle={styles.imageOpacity}>
+                <Text style={{ ...styles.imageText, ...styles.imageTextMain }}>{item[imageMainData]}+</Text>
```
arribbar, line 41: Let's do the same that for @guillaumemoby, you can use destructuration for all styles juste above the `return`
The same for item.
For the renderItem, you van even have a new function outside the render function

src/Components/DrillListPage.js
```diff
@@ -18,6 +19,9 @@ export const DrillListPage = props => {
   const [data, setData] = useState(drills);
   const [displayFilters, setDisplayFilters] = useState(false);
 
+  const imageMainData = type === DrillTypes.TECHNICAL ? 'nbPlayers' : 'durationInMinutes';
+  const imageMainDataLegend = type === DrillTypes.TECHNICAL ? 'players' : 'min.';
```
arribbar, line 23: It looks like it is always technical on my device for technical and fitness drills ... Is it normal ?

sallesma: Yes I made a mistake on the home page when using enums. I fix it

src/Components/DrillListPage.js
sallesma: I'm not sure at all it is considered a good practice to destructure styles. There are many of them and their names could be misleading once declared in the component namespace.

I can't say i'm an experienced react-native developer but I haven't seen this done anywhere?

src/Components/DrillListPage.js
sallesma: Good that I wrote tests on navigation in the other PR :D

src/Components/DrillListPage.js
sallesma: I went for information on the web and only found this article : https://codereview.stackexchange.com/questions/165936/using-a-destructuring-assignment-on-a-style-object-to-extract-two-members

Keep ieas seem to be :
- "I think it is a matter of preference"
- "You could also compromise and go for s = styles"
- "Also, body or button have unclear semantics without the style. prefix "

I think I would stick to calling `styles.something`

> Clarity and brevity sometimes are at odds. When they are, I choose clarity. 
> -- Jacob Kaplan-Moss

src/Components/DrillListPage.js
```diff
@@ -33,13 +37,14 @@ export const DrillListPage = props => {
...
-              <Image style={styles.image} source={{ uri: item.image }} />
+              <ImageBackground source={{ uri: item.image }} style={styles.image} imageStyle={styles.imageOpacity}>
+                <Text style={{ ...styles.imageText, ...styles.imageTextMain }}>{item[imageMainData]}+</Text>
+                <Text style={styles.imageText}>{imageMainDataLegend}</Text>
+              </ImageBackground>
               <View style={styles.contentContainer}>
                 <Text style={styles.source}>{item.source}</Text>
                 <Text style={styles.title}>{item.title}</Text>
-                <Text style={styles.numberOfPlayers}>
-                  Duration: {item.durationInMinutes} min - players: {item.nbPlayers}
```
Adeloche, line 41: So we don't display drill duration anymore when it is a technical drill ?

sallesma: Nope, because yesterday we decided drill duration is not so important when talking about a technical drill (compared with fitness where duration is one of the most important piece of info)

But it's going to be displayed on the drill page.
Also, probably we should display it on the training page with a small drill list, since timing is important there

Adeloche: Makes sense :)

src/Components/DrillListPage.js
arribbar: Ahah, that's good for me

src/Components/DrillListPage.js
arribbar: Ahaha

PR #55, opened 2020-04-04
Training List page

Adeloche:
Partly solves #19 .

This is page groups together sessions that we will have stored (sessions are a list of drills).

To be implemented in a further PR: 
- filter for sessions (I was not sure what kind of filters we want here)

Open questions:
- Do we need to display the training sessions 'source' as it is done for drills ?
- I assumed that the sessions duration was the sum of duration of the drills into it, and le nb of players was the max of all nb players required by drills into it. I thought computing them instead of having to fill them manually would make easier training edition. Is this ok ?
- I need some help to implement test of this page I think

here is what it looks like:

image


src/Components/TrainingListPage.js
```diff
@@ -0,0 +1,101 @@
+import React, { useState } from 'react';
+import { StyleSheet, View, Text, Image, TouchableOpacity, FlatList } from 'react-native';
+import { connect } from 'react-redux';
+import theme from '../styles/theme.style';
+
+const getTrainingDuration = trainingDrills => {
+  const durationList = trainingDrills.map(({ durationInMinutes }) => durationInMinutes);
+  const totalDurationInMinutes = durationList.reduce((a, b) => a + b, 0);
```
sallesma, line 8: Maybe we can make it simpler with something oneline like `return trainingDrills.reduce((a, b) => a.durationInMinutes + b.durationInMinutes, 0);` ?

src/Components/TrainingListPage.js
```diff
@@ -0,0 +1,101 @@
...
+  return totalDurationInMinutes;
+};
+
+const getTrainingNbPlayers = trainingDrills => {
+  const nbPlayersList = trainingDrills.map(({ nbPlayers }) => nbPlayers);
+  const totalNbPlayers = Math.max(...nbPlayersList);
```
sallesma, line 14: Here you could return directly the result of Math.max(...) without assigning it to a const

src/Components/TrainingListPage.js
```diff
@@ -0,0 +1,101 @@
...
+      <Text style={styles.counter}>{data.length} training sessions available</Text>
+      <FlatList
+        data={data}
+        keyExtractor={item => item.id.toString()}
+        renderItem={({ item }) => (
+          <TouchableOpacity
+            style={styles.training}
+            onPress={() => navigation.navigate('TrainingPage', { training: item })}
+          >
+            <Image style={styles.image} source={{ uri: item.image }} />
```
sallesma, line 41: If you have a look at #54 you could implement the same modification on image opacity

src/Components/TrainingListPage.js
```diff
@@ -0,0 +1,101 @@
...
+          </TouchableOpacity>
+        )}
+      />
+    </View>
+  );
+};
+
+export default connect(mapStateToProps)(TrainingListPage);
+
+const styles = StyleSheet.create({
```
sallesma, line 59: It seems styles are copy/pasted for different lists. Maybe we should extract them to a separate file and use the same styles on both DrillListPage and TrainingPage ?

Adeloche:
> Cool!
> I can help you with the tests if needed. I have time tomorrow :)

I did not think I would but it looks like a managed to get something working. :)
I'll try write a short how to tomorrow

src/Components/TrainingListPage.js
Adeloche: Sounds good. What do you think is the relevant information for training sessions ? Duration or nb player ?

sallesma: I have no idea 🤷‍♂ 
Maybe we should let it empty here? The simpler the better

src/Components/TrainingListPage.js
Adeloche: ok, done

src/Components/TrainingListPage.js
Adeloche: So I tried, but without anything written on the image the result is kind of ugly.
Can I show you tomorrow so we can decide what to do ?

Adeloche: In the end I put both information. I like the idea of using the image space.
I guess we can still change later if we think only part of the information (or none) is relevant.

src/Components/TrainingListPage.js
```diff
@@ -0,0 +1,95 @@
...
+  };
+};
+
+export const TrainingListPage = props => {
+  const { allDrills, trainings, navigation } = props;
+
+  const [data] = useState(trainings);
+
+  return (
+    <View style={styles.trainingListPage}>
```
arribbar, line 30: You should use destructuration for `styles` and `item` as well :)
It's easier like this to know what the component is using

arribbar:
Ok, my bad, for `style` and destructuration, you can keep it like this.
For `item` and destructuration, I would say it depends if you repeat a lot `item.yourProp`. 

But ... Now, that I have merged #54 , there is one conflict to resolve before merging

src/Components/DrillListPage.js
```diff
@@ -64,39 +66,25 @@ const styles = StyleSheet.create({
     height: '100%',
   },
   counter: {
-    color: theme.COLOR_SECONDARY,
-    marginBottom: 20,
+    ...list.counter,
```
sallesma, line 69: I think you don't even need to declare a counter style here.
You could use directly `list.counter` on your element since we don't change anything to it
That will reduce the boilerplate code a lot

(same for the others)

src/Components/TrainingListPage.js
```diff
@@ -0,0 +1,95 @@
...
+
+const styles = StyleSheet.create({
+  trainingListPage: {
+    paddingTop: 10,
+    paddingLeft: 20,
+    backgroundColor: theme.BACKGROUND_COLOR_LIGHT,
+    height: '100%',
+  },
+  counter: {
+    ...list.counter,
```
sallesma, line 69: Same here you can use directly the styles from list without redeclaring them if you don't override anything

src/Components/TrainingListPage.test.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+import { Provider } from 'react-redux';
+import store from '../Store/configureStore';
+import fixtures from '../Fixtures/TestFixtures';
+
+import ConnectedTrainingListPage, { TrainingListPage } from './TrainingListPage';
+
+const storeInstance = store;
```
sallesma, line 9: You can remove this line and use `store` directly on the test below

(That useless line is from me and I removed it from the other existing test in this PR #47 )

Adeloche:
> Ok, my bad, for `style` and destructuration, you can keep it like this.
> For `item` and destructuration, I would say it depends if you repeat a lot `item.yourProp`.
> 
> But ... Now, that I have merged #54 , there is one conflict to resolve before merging

I fixed the conflict. However, I have no idea of what you mean by tem destructuration. Can you explain this to me or send me a useful article explaining this ?
Sorry it is my first project with React

src/Components/TrainingListPage.js
```diff
@@ -0,0 +1,68 @@
...
+
+const styles = StyleSheet.create({
+  trainingListPage: {
+    paddingTop: 10,
+    paddingLeft: 20,
+    backgroundColor: theme.BACKGROUND_COLOR_LIGHT,
+    height: '100%',
+  },
+  contentContainer: {
+    ...list.contentContainer,
```
sallesma, line 66: Is there a reason to keep this one here?

src/Components/TrainingListPage.js
```diff
@@ -0,0 +1,68 @@
...
+  const [data] = useState(trainings);
+
+  return (
+    <View style={styles.trainingListPage}>
+      <Text style={list.counter}>{data.length} training sessions available</Text>
+      <FlatList
+        data={data}
+        keyExtractor={item => item.id.toString()}
+        renderItem={({ item }) => (
+          <TouchableOpacity style={list.item} onPress={() => navigation.navigate('TrainingPage', { training: item })}>
```
sallesma, line 36: I guess you could add a test on navigation (there should be an example now in DrillListPage.test (the debug variable is probably useless there btw)

arribbar:
> About destructuration, if you have an item object like this :
> 
> ```
> const item = {
>   name: "Hello",
>   description: "blah blah blah",
> }
> ```
> 
> And you do several times `item.name` in the same component
> You can use `const {name, description} = item` and then use directly `name` and `description`
> 
> A lot of people like to destructure `props` at the beginning of the render method so that is is very clear what props (= what dependencies) the render method is using.
> 
> Going back to our code, it is not easy to do this with our item since it is nested, but if we extracted a component or function to render each item, we could destructure the item object to be explicit on what this component/function is using.
> I am not sure but I think that's what @arribbar means

Yep, that how you do destructuration, thanks @sallesma 
There are two things to be careful to keep your codebase  more readable:  
-  Write short functions each time you can do it. Here you can do it for `renderItem` but if you think it is still readable, you can let it like this. 
-  Use destructuration. As said @sallesma, it's really helpful if you use several times properties in `item` for this case. So If we take `renderItem` you can do it in two ways, either in another function or directly within doing:
```
renderItem={({ item }) => {
const { 
prop1,
prop2,
} = item;
return (
<TouchableOpacity ... 
```
If you are not familiar with es6, I suggest you to take a look at some [cheat sheets] (https://mbeaudru.github.io/modern-js-cheatsheet/). It would save you a lot of time and used a lot in React (and most web browsers can read es6 by now) ;)

src/Components/TrainingListPage.js
Adeloche: as disscussed together, will do in #26

src/Components/TrainingListPage.js
Adeloche: no, I just forgot :)

src/Components/TrainingListPage.js
Adeloche: done

Adeloche:
> > About destructuration, if you have an item object like this :
> > ```
> > const item = {
> >   name: "Hello",
> >   description: "blah blah blah",
> > }
> > ```
> > 
> > 
> > And you do several times `item.name` in the same component
> > You can use `const {name, description} = item` and then use directly `name` and `description`
> > A lot of people like to destructure `props` at the beginning of the render method so that is is very clear what props (= what dependencies) the render method is using.
> > Going back to our code, it is not easy to do this with our item since it is nested, but if we extracted a component or function to render each item, we could destructure the item object to be explicit on what this component/function is using.
> > I am not sure but I think that's what @arribbar means
> 
> Yep, that how you do destructuration, thanks @sallesma
> There are two things to be careful to keep your codebase more readable:
> 
> * Write short functions each time you can do it. Here you can do it for `renderItem` but if you think it is still readable, you can let it like this.
> * Use destructuration. As said @sallesma, it's really helpful if you use several times properties in `item` for this case. So If we take `renderItem` you can do it in two ways, either in another function or directly within doing:
> 
> ```
> renderItem={({ item }) => {
> const { 
> prop1,
> prop2,
> } = item;
> return (
> <TouchableOpacity ... 
> ```
> 
> If you are not familiar with es6, I suggest you to take a look at some [cheat sheets](https://mbeaudru.github.io/modern-js-cheatsheet/). It would save you a lot of time and used a lot in React (and most web browsers can read es6 by now) ;)

as discussed, will do in PR #26

src/Components/TrainingListPage.js
sallesma: Thanks

PR #68, opened 2020-04-07
Upgrade expo

### LINKED ISSUE #57: Security upgrades
We have security alerts on Mem and Minimist

Minimist alert (high severity) come from a dependency of a dependency of a dependency of Jest.

So we should update jest

We could also set up an automatic update of npm dependencies. Otherwise we'll end up outdated after a while

sallesma:

Closes #57


sallesma:
That is the result from `expo upgrade`

PR #69, opened 2020-04-10
Optimize app loading

sallesma:

### DIFF
--- a/App.js
+++ b/App.js
@@ -1,9 +1,11 @@
 import 'react-native-gesture-handler';
-
 import React from 'react';
-import { Navigation } from './src/Navigation';
+import { enableScreens } from 'react-native-screens';
 import { Provider } from 'react-redux';
+import { Navigation } from './src/Navigation';
 import Store from './src/Store/configureStore';
+
+enableScreens();
 
 export default class App extends React.Component {
   render() {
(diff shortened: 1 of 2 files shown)


App.js
```diff
@@ -1,16 +1,51 @@
...
 import Store from './src/Store/configureStore';
 
-export default class App extends React.Component {
-  render() {
-    return (
-      <Provider store={Store}>
-        <Navigation />
-      </Provider>
-    );
+enableScreens();
```
sallesma, line 10: This is recommended by expo : https://reactnavigation.org/docs/react-native-screens

App.js
```diff
@@ -1,16 +1,51 @@
...
+
+const PERSISTENCE_KEY = 'NAVIGATION_STATE';
```
sallesma, line 12: Everything else is from this rect-navigation guide : https://reactnavigation.org/docs/state-persistence

If you think it's not useful for our users, we could also activate it only on development. It's pretty annoying to go back to the home page on reload

src/Navigation/Navigation.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { NavigationContainer } from '@react-navigation/native';
+import { AsyncStorage } from 'react-native';
```
arribbar, line 2: It seems you are not using `AyncStorage` in this file

sallesma: 👍

PR #74, opened 2020-04-17
Add real drills and trainings

sallesma:

I didnt spend to much time on explanation. The animation will follow when available

Looks like this at the moment :

Screenshot 2020-04-17 at 23 55 29

### DIFF
--- a/src/Components/TrainingPage.js
+++ b/src/Components/TrainingPage.js
@@ -23,7 +23,7 @@ export const TrainingPage = props => {
   const trainingDrills = training.drills.map(drillId => drills.find(drill => drill.id === drillId));
 
   return (
-    <ScrollView style={styles.trainingPage}>
+    <ScrollView style={styles.trainingPage} contentContainerStyle={styles.trainingPageContent}>
       <View style={styles.infos}>
         <Text style={styles.descriptionText}>{training.description}</Text>
         <View style={styles.infoDisplay}>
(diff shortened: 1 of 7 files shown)


src/Components/TrainingPage.js
```diff
@@ -9,74 +9,82 @@ export function getGoals(trainingDrills) {
...
-export class TrainingPage extends Component {
-  render() {
-    const { navigation } = this.props;
-    const training = this.props.route.params.training;
-    const drills = training.drills.map(drillId => this.props.drills.find(drill => drill.id === drillId));
-
-    return (
-      <View style={styles.trainingPage}>
+  return (
+    <ScrollView style={styles.trainingPage} contentContainerStyle={styles.trainingPageContent}>
```
Adeloche, line 26: Don't you get a warning using a FlatList inside this scrollview ? ?

sallesma: Ah, yes I got the warning. It seems to be related to performance issue on large lists of elements and I don't think we'll have more than 10 drills there.

But I'll look for a solution tomorrow

sallesma: I had a look at it this morning and it seems that:
- It is a performance issue because all the elements are rendered even if they are not visible on the screen. This is not so important here since training drills are < 10
- VDD is going to think about the training page and probably going to suggest something pretty different, so w'll have to refactor it later on anyways

PR #75, opened 2020-04-18
Newdrillpage

arribbar:


src/Components/DrillAnimationPage.js
```diff
@@ -1,18 +1,16 @@
 import React, { Component } from 'react';
-import { StyleSheet, View, Text } from 'react-native';
+import { StyleSheet, View, Text, Dimensions } from 'react-native';
 import Animation from './Animation';
 import { WebView } from 'react-native-webview';
 
-class DrillPage extends Component {
+class DrillAnimationPage extends Component {
```
sallesma, line 6: Maybe we could just call it `DrillAnimation` if it's not a page anymore

src/Components/DrillAnimationPage.js
```diff
@@ -1,18 +1,16 @@
...
   render() {
-    const drill = this.props.route.params.drill;
-
     return (
       <View style={styles.drillAnimationPage}>
-        {drill.animation ? (
-          <Animation animation={drill.animation} />
-        ) : drill.video ? (
-          <WebView source={{ uri: drill.video }} style={styles.videoPresentation} />
+        {this.props.animation ? (
```
sallesma, line 10: Maybe extract the props at the beginning of the render method ?

And switch to a functional component... Or even remove this componentt ? It does nothing but a `if/else`. It could be done directly on DrillPage

package.json
```diff
@@ -27,6 +27,7 @@
     "react-native-screens": "~2.2.0",
     "react-native-web": "^0.11.7",
     "react-native-webview": "8.1.1",
+    "expo-linear-gradient": "~8.0.0",
```
sallesma, line 30: I have a warning 

```
Some of your project's dependencies are not compatible with currently installed expo package version:
 - expo-linear-gradient - expected version range: ~8.1.0 - actual version installed: ~8.0.0
Your project may not work correctly until you install the correct versions of the packages.
To install the correct versions of these packages, please run: expo install [package-name ..
```

We should use the expo-compliant version

src/Components/DrillPage.js
```diff
@@ -140,9 +199,50 @@ const styles = StyleSheet.create({
...
+    color: theme.COLOR_SECONDARY,
+  },
+  descriptionGoalsText: {
+    color: theme.COLOR_PRIMARY,
+    textAlign: 'center',
+    fontSize: theme.FONT_SIZE_LARGE,
+    fontWeight: 'bold',
+  },
+  lines: {
+    borderBottomColor: '#DCDCDC',
```
sallesma, line 240: Colors should be part of the theme.
Maybe use an already existing one that is close, like `COLOR_SECONDARY_LIGHT` (e5e5e5)

src/Components/DrillPage.js
```diff
@@ -103,34 +165,31 @@ const styles = StyleSheet.create({
...
     borderRightWidth: 1,
     borderRightColor: theme.COLOR_PRIMARY_LIGHT,
   },
   videoLink: {
-    margin: 40,
     width: 100,
     height: 100,
     borderRadius: 50,
-    backgroundColor: theme.COLOR_PRIMARY_LIGHT,
+    backgroundColor: '#FFF933',
```
sallesma, line 176: It looks like this color is useless since we use a linear gradient that defines colors

src/Components/DrillAnimationPage.js
guillaumemoby: We are going to change it in the branch vimeovideo. Could we keep it ?

sallesma: Yep 👍

PR #76, opened 2020-04-19
Better filters ui

sallesma:

## Main changes

- Buttons are more regular
- goals are displayed as checkbox
- use a new library for slider since the old one is being deprecated, and the checkbox library also provides slider
- remove "Anytime" from season timings filters since it is equivalent to no filtering
- extract the three filters components (button, checkbox, slider) to a dedicated `filters`folder

## Before

Screenshot 2020-04-19 at 18 50 34

## After

Screenshot 2020-04-19 at 19 04 10


src/Components/FitnessFilters.js
```diff
@@ -145,12 +149,11 @@ class FitnessFilters extends React.Component {
           </View>
           <Text style={filterStyle.filterTitle}>Duration: {durationInMinutes || '-'} mins</Text>
           <Slider
-            minimumValue={1}
+            minimumValue={10}
```
Adeloche, line 152: maybe use dynamic value for min and max ? like max and min value of available drills ?

sallesma:
> Can we allow to select multiple values in filter ?

Would be better I think, I'll try to think about it. If I can't come up with a small change for this, I'll merge this PR and then work on another branch to make it easy to review

> The 'Duration' slider looks nice but I'm not sure the meaning of this filter is very clear.

Yes it feels like a dumb robot asking a stupid question "duration ?"

If we keep using the slider, the question should be more clear. eg : "How much time do you have?"
If we move to buttons (10, 20, 30 min) it stays unclear to me what "10 mins" means : is it drills from 10min long, or drills upto 10 mins long?

Si I guess the most important is the question. And I think as a coach using the app, I want to filter by the time I want to spend on the exercise, so the question could be "How much time do you have?" on fitness and "How many players do you have?" on technical

What do you think?

src/Components/FitnessFilters.js
sallesma: I think depends on the question the user is answering like you said in your message above

Adeloche:
> > Can we allow to select multiple values in filter ?
> 
> Would be better I think, I'll try to think about it. If I can't come up with a small change for this, I'll merge this PR and then work on another branch to make it easy to review
> 
Yes !
> > The 'Duration' slider looks nice but I'm not sure the meaning of this filter is very clear.
> 
> Yes it feels like a dumb robot asking a stupid question "duration ?"
> 
> If we keep using the slider, the question should be more clear. eg : "How much time do you have?"
> If we move to buttons (10, 20, 30 min) it stays unclear to me what "10 mins" means : is it drills from 10min long, or drills upto 10 mins long?
> 
> Si I guess the most important is the question. And I think as a coach using the app, I want to filter by the time I want to spend on the exercise, so the question could be "How much time do you have?" on fitness and "How many players do you have?" on technical
> 
> What do you think?

Updating the questions like you suggested sounds good for a first version. I think though, at least for fitness, that when I have 30 minutes, I'm not very interested in seing the 5 minutes drills. But maybe this detail won't bother  our users, anyway I think we can wait and see how the app is used.

sallesma:
I gave it a try and updated the slider so that it's more explicit

Screenshot 2020-04-23 at 00 05 07

I think it's ok for now, we'll probably change it later, maybe we'll need a two-marker slider with min and max

Also fixed the warning @arribbar pointed out

PR #78, opened 2020-04-22
Vimeovideo

arribbar:

**Done**
- Add a `source` between `'vimeo`, `youtube` and `animation` to a step
- Display a video from vimeo
- Play a list of steps (only videos for now) with a finish at the end

**Question**
My friend @guillaumemoby changed a bit how was done DrillAnimation and is now more a PlayerStepList. I am not a big fan of how it is now.
I would suggest to split this file in two, one as PlayerStepList and the other one `PlayStep` (with animation, video or webview)
What do you think about this proposal?


package.json
```diff
@@ -27,10 +29,10 @@
     "react-native-screens": "~2.2.0",
     "react-native-web": "^0.11.7",
     "react-native-webview": "8.1.1",
-    "expo-linear-gradient": "~8.1.0",
     "react-redux": "^7.2.0",
     "redux": "^4.0.5",
-    "watchman": "^1.0.0"
+    "watchman": "^1.0.0",
+    "expo-screen-orientation": "~1.0.0"
```
sallesma, line 35: Did you add this lib with expo install ?

src/Fixtures/index.js
```diff
@@ -53,18 +53,22 @@ export const EquipmentLabels = {
   FULL: 'Full equipment',
 };
 
+export const Sources = {
```
sallesma, line 56: `VideoSources` ?

src/Components/VimeoVideo.js
```diff
@@ -0,0 +1,51 @@
...
+            });
+            await videoElem.current.setStatusAsync({
+              rate: 1.0,
+              isMuted: true,
+              resizeMode: Video.RESIZE_MODE_CONTAIN,
+              shouldPlay: true,
+              isLooping: true,
+            });
+          } catch (e) {
+            console.log('ERROR Loading Video', e);
```
sallesma, line 35: Should we provide user feedback ?

src/Components/VimeoVideo.js
```diff
@@ -0,0 +1,51 @@
+import React, { useEffect, useRef } from 'react';
+
+import { Video } from 'expo-av';
+
+const VimeoVideo = props => {
+  const vimeoUrlSource = `https://player.vimeo.com/video/${props.vimeoId}/config`;
+  const videoElem = useRef(null);
+
+  useEffect(() => {
+    async function fetchVimeoData() {
```
sallesma, line 10: Why do you need to define a function and call it just afterwards ?

src/Components/VimeoVideo.js
```diff
@@ -0,0 +1,51 @@
...
+            });
+          } catch (e) {
+            console.log('ERROR Loading Video', e);
+          }
+        });
+    }
+    fetchVimeoData();
+
+    return () => {
+      console.log('Error loading video content');
```
sallesma, line 42: So this is called when the component unmounts (if my knowledge of useEffect is ok), I don't see why we do this

src/Components/DrillAnimation.js
```diff
@@ -1,29 +1,245 @@
...
+    borderRightColor: theme.COLOR_PRIMARY_LIGHT,
+  },
+  buttonNext: {
+    margin: 22.5,
+    width: 25,
+    height: 25,
+    borderRadius: 12.5,
+    backgroundColor: '#FFF',
+    borderWidth: 2,
+    borderColor: '#808080',
```
sallesma, line 181: We should use colors from the theme

src/Components/DrillAnimation.js
```diff
@@ -1,29 +1,245 @@
...
+          </View>
+          <View style={styles.lines} />
+        </View>
+      );
+    } else {
+      return (
+        <View style={styles.infoWrapper}>
+          <View style={styles.description}>
+            <View style={styles.subWrapper}>
+              <Text style={styles.fitnessNext}>{props.drill.steps[count + 1].count}</Text>
```
sallesma, line 34: Maybe it would be easier to understand if count was called `currentStepIndex` ?

src/Components/DrillAnimation.js
```diff
@@ -1,29 +1,245 @@
...
+      }
+    }
+  };
+
+  const displayFinish = () => {
+    return (
+      <View>
+        <View style={styles.containerFinish}>
+          <View style={styles.infoWrapper}>
+            <TouchableOpacity style={styles.buttonFinish} onPress={() => incrementCount()}>
```
sallesma, line 72: Maybe we should extract a button wince we use it on several pages and we are probably going to use it as well on the trainings

We already have a HeaderButton and a FilterButton. Is this one a `CtaButton`. I don't have a lot of ideas 😕

src/Components/DrillAnimation.js
```diff
@@ -1,29 +1,245 @@
...
+      );
+    }
+  };
+
+  const checkSwitch = () => {
+    if (props.drill.steps && count === props.drill.steps.length) {
+      return displayFinish();
+    } else {
+      switch (props.drill.steps && props.drill.steps[count].source) {
+        case 'animation':
```
sallesma, line 52: Use the enum here ?

src/Components/DrillAnimation.js
```diff
@@ -1,29 +1,245 @@
...
+        </View>
+      </ScrollView>
+    );
+  };
+
+  const displayYoutube = () => {
+    return (
+      <WebView
+        source={{
+          uri: props.drill.steps[count].link,
```
Adeloche, line 109: What do you think of storing the youtube ID of the video in the step instead of full url ? And call the full URL in this displayYoutube function.
Depending on the used url (for example if embed option in included in url), the video display can be different, I think we want to be sure the display is always the same.

package.json
arribbar: Nope, and this is @guillaumemoby's work ^^

arribbar: Well, I don't know

arribbar: Bu I'll remove it as we don't use it for now

arribbar: Done

src/Fixtures/index.js
arribbar: Hmm, don't know if it is better as it is between `animation`, `youtube` and `vimeo`... `sourceType` ?

src/Components/VimeoVideo.js
arribbar: It could be a good idea but the question is how do we provide a feedback ?
We could also indicate the video is loading as well

src/Components/VimeoVideo.js
arribbar: Because it is an `async` function and so return a `Promise`.
And a Hook should return nothing or a cleanup function

src/Components/VimeoVideo.js
arribbar: Don't know why neither ^^

arribbar: done

src/Components/DrillAnimation.js
arribbar: Done

src/Components/DrillAnimation.js
arribbar: Done

src/Components/DrillAnimation.js
arribbar: I named it `GradientButton` ;)

src/Components/DrillAnimation.js
arribbar: Done

src/Components/DrillAnimation.js
arribbar: It could be a nice idea.
Depends on how we store the data indeed

arribbar:
> Nice !
> I'm afraid I haven't understood your question about PlayerStep/PlayStepList

All right, before we were just displaying depending on the source either an animation, a video from youtube or a video from vimeo
Now, we are displaying a list of steps with for each step a comment, a next button and what we were displaying before, the animation, youtube video or vimeo video

I'm gonna split this file in two to keep them separated :)

arribbar:
Hmm, in fact, I am quite undecided ...
It really depends on how we structure our data. 
Can an exercise (or a drill) have several steps ? 
Or it is only a training with several exercises ?

arribbar:
Ok, so depending how we structure our data, I'll need to change a bit our code.
Either in `Fixture/index.js` and/or DrillAnimation.js
Also, I have still a problem testing the DrillPage @sallesma

src/Fixtures/index.js
sallesma: It's just that `Source` is so generic I would exect it to be the source of the drill, like `UltimateApp`, `RiseUp`, `Other` (close to "author")

Suggestions : `DisplayType`, `IllustrationType` ?

src/Components/DrillAnimation.js
```diff
@@ -68,64 +66,55 @@ const DrillAnimation = props => {
     return (
       <View>
```
sallesma, line 67: Do we need that many View components ?

src/Components/DrillAnimation.js
```diff
@@ -1,23 +1,24 @@
...
 import { WebView } from 'react-native-webview';
-import { LinearGradient } from 'expo-linear-gradient';
 
 import Animation from './Animation';
 import VimeoVideo from './VimeoVideo';
+import GradientButton from './shared/GradientButton';
+import { Sources } from '../Fixtures';
 import theme from '../styles/theme.style';
 
 const DrillAnimation = props => {
```
sallesma, line 11: Following my question on the `SourceType` enum, I think the same questions pops out here : We use the word Animation for two purposes and it makes it more difficult to follow

DrillAnimation : Animation or video
Animation : Animated diagram

Whe should go with different words for the two concepts

Suggestion :
- DrillIllustration : animation or vidéo
Animation : animated Diagram
- DrillAnimation : Diagram or video
Diagram : animated diagram

What do you think?

src/Components/VimeoVideo.js
sallesma: Maybe we could remove the async/await and just call fetch
Then we could display something while waiting for the video to load ?

src/Components/DrillAnimation.js
arribbar: I agree with you. I would remove the word Animation as it could be anything.
So, I would choose 
- `DrillIllustration` : animation or video
- `Diagram`: animated diagram

Is it ok for you ?

src/Fixtures/index.js
arribbar: Ah, yes, agree with you
Let's go for `IllustrationType`

arribbar: I choose `IllustrationType`
By the way, do you think we change the naming in steps (`sourceType` and `source`) or no ?
And I renamed `source` in `drill` by `author`

arribbar:
I remove the `async` in the `fetch`
And added a loader when video is loading. You can at least see it where you click on the button `Do it Again`
By the way, I wanted to add some tests to vimeoVideo but, I do not know how to do as we need to do a fetch to get the video url ...

src/Components/DrillAnimation.js
sallesma: Yep

sallesma: Maybe we should ask on Slack if everyone is ok with a renaming Animation->Diagram, since it will cause conflicts ?

src/Fixtures/index.js
sallesma: >  By the way, do you think we change the naming in steps (sourceType and source) or no ?

Probalby yes, to be consistent?

src/Components/DrillAnimation.js
sallesma: Much better than `CtaButton` 😄

src/Components/DrillAnimation.js
sallesma: I like it : the data would be responsible for "which video" and the component for "how to display it"

knowing the full url is robably more about "how"

src/Components/VimeoVideo.test.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+
+import VimeoVideo from './VimeoVideo';
+import { VIMEO_VIDEO_ID } from '../Fixtures';
+
+describe('<VimeoVideo />', () => {
+  it('renders correctly for a video from vimeo', () => {
```
sallesma, line 8: I'm going to give it a try

I found this library which seems to do the job and be widely used

Going to use the recording feature to generate the mocks since video content can be pretty awful to write : https://github.com/nock/nock#recording
Then going to use the mocks
Then going to disable network requests for the whole test suite to make sure any call to the outer world is mocked : https://github.com/nock/nock#enabledisable-real-http-requests

sallesma: It looks like my test works, except on node.js v12...

src/Components/DrillAnimation.js
arribbar: Yep, I'll suggest to do it in another PR

src/Components/DrillAnimation.js
arribbar: Sure. And I don't think the best option is to do it but better with the Zach's PR

src/Components/VimeoVideo.test.js
sallesma: Looks ok now
Can someone review the tests I wrote?

PR #79, opened 2020-04-23
Introduce minimal drillpage to use in trainings

aenario:

This can be merged as-is, althought styling could use some improvement, but I am waiting on the Animation PR to land before making changes to the animation component so it can be rendered at a smaller size.

|Screenshot_20200423-195002   | Screenshot_20200423-195102  |
|---|---|


src/Components/DrillPageMinimal.js
```diff
@@ -0,0 +1,159 @@
+import React, { useLayoutEffect } from 'react';
+import { ScrollView, StyleSheet, View, Text, Image, TouchableOpacity } from 'react-native';
+import { connect } from 'react-redux';
+
+import Progress from './ProgressBar2';
+import DrillAnimation from './DrillAnimation';
+
+import iconCheckMark from '../Images/ic_check.png';
```
sallesma, line 8: We have an asset folder, maybe you could move this one there as well

src/Components/DrillPageMinimal.js
```diff
@@ -0,0 +1,159 @@
...
+
+  const goToFullDrill = () => {
+    navigation.navigate('DrillPage', { drill });
+  };
+
+  useLayoutEffect(() => {
+    navigation.setOptions({
+      headerRight: () => {
+        return (
+          <TouchableOpacity
```
sallesma, line 40: There is a HeaderButton inside shared folder that does exactly the same thing.

You could use it here (and tweak it if you want it to use the iconCheckMark image)

src/Components/ProgressBar2.js
```diff
@@ -0,0 +1,94 @@
+import React, { useRef, useEffect } from 'react';
+import { StyleSheet, Animated, View, TouchableOpacity } from 'react-native';
+
+/** Progress bar displayed with an animation which:
```
sallesma, line 4: This comment is a bit confusing since you use it to display the progression within the training. Maybe we should remove it?

src/Components/ProgressBar2.js
```diff
@@ -0,0 +1,94 @@
...
+const useAnimation = targetValue => {
+  const animatedValue = useRef(new Animated.Value(targetValue)).current;
+  const startAnimation = () => {
+    Animated.timing(animatedValue, { toValue: targetValue }).start();
+  };
+  useEffect(startAnimation, [targetValue]);
+  return animatedValue;
+};
+
+export const ProgressBar = props => {
```
sallesma, line 28: We could add a basic test for it (at least snapshot test)

src/Components/ProgressBar2.js
```diff
@@ -0,0 +1,94 @@
...
+    outputRange: ['0%', '100%'],
+  });
+
+  return (
+    <View style={styles.progressContainer}>
+      <View style={styles.progressInner}>
+        <Animated.View style={styles.lineBackground} />
+        <Animated.View style={[styles.lineFiller, { width }]} />
+
+        {range(1, total).map(k => (
```
sallesma, line 42: Is there a reason why you use the letter k? Is it a convention?

src/Components/DrillPageMinimal.js
```diff
@@ -0,0 +1,159 @@
...
+    drills: state.drills,
+  };
+};
+
+export const DrillPageMinimal = props => {
+  const { route, navigation, drills } = props;
+  const { drill, training } = route.params;
+
+  const totalDrills = training.drills.length;
+  const currentDrill = training.drills.findIndex(id => id === drill.id);
```
sallesma, line 23: Could we call it `currentDrillIndex` or something similar?

src/Components/DrillPageMinimal.js
```diff
@@ -0,0 +1,159 @@
...
+
+  return (
+    <ScrollView style={styles.drillPage}>
+      <View style={styles.titleContainer}>
+        <Text style={styles.title}>
+          Drill {currentDrill + 1} - {drill.title}
+        </Text>
+        <Progress total={totalDrills} current={currentDrill + 1} onDotPress={onProgressDotPress} />
+      </View>
+      <View style={styles.description}>
```
sallesma, line 66: Reading the component, I felt like everything below this comment is the `MinimalDrill` (its responsability is to display attributes of the drill) and the training management could be a `LiveTrainingPage` that displays basic training info (title/progressbar) and manages a `currentDrill` that is passed as a props to `MinimalDrill`.

From the wireframe, it seems we'd like to add what I call `MinimalDrill` to our TrainingPage as well

Not sure if my comment is clear... What do you think?



Screenshot 2020-04-24 at 21 23 10

src/Components/ProgressBar2.js
aenario: k is for key, but key will be more explicit, will fix.

src/Components/DrillPageMinimal.js
aenario: Good idea. On it

aenario:
Rebased on master

package.json
```diff
@@ -15,8 +15,9 @@
   "dependencies": {
     "@react-native-community/masked-view": "0.1.6",
     "@react-navigation/native": "^5.1.1",
+    "@react-navigation/routers": "^5.3.0",
```
sallesma, line 18: Wondering what this is used for?

I can't find any usage in this pr

src/Components/DrillPageMinimal.js
```diff
@@ -0,0 +1,123 @@
...
+    alignItems: 'center',
+  },
+  btnMore: {
+    width: '80%',
+    backgroundColor: theme.COLOR_PRIMARY,
+    textAlign: 'center',
+    padding: 20,
+  },
+  btnMoreContent: {
+    color: 'rgb(255, 255, 255)',
```
sallesma, line 121: Maybe use a color from the theme ?

aenario: I need to do a whole lot of styling on this view later on, but i believe we should merge as-is to have the full navigation flow and avoid conflicts on the training view.

package.json
aenario: It's a transitive dependency of react-navigation/native, but i needed it to be the latest version to avoid errors with `expo --web`

```
UltimateApp
└─┬ @react-navigation/native@5.1.4
  └─┬ @react-navigation/core@5.3.2
    └── @react-navigation/routers@5.2.1 
```

src/Components/DrillPageMinimal.js
sallesma: Yep, that was a non-blocking comment for me :)

PR #80, opened 2020-04-24
Programs

sallesma:

Screenshot 2020-04-24 at 22 55 00

The UI is OK, we can make it better later

The real completion of programs should be implemented in another PR because it will introduce a lot of complexity in our redux store

Most of the diff comes from the fact that now, trainings contain drill objects instead of an id (same for programs and their trainings). It makes navigation easier since we don't have to look through the whole store each time we want to find something


src/Components/ProgramsPage.js
```diff
@@ -0,0 +1,37 @@
...
+  return (
+    <View style={styles.programPage}>
+      <FlatList
+        data={programs}
+        keyExtractor={item => item.id.toString()}
+        renderItem={({ item }) => <Program program={item} navigation={navigation} />}
+      />
+    </View>
+  );
+};
```
Adeloche, line 20: If I remember well, we  decided to keep a link to a full trainings list as well, below the predefined program.

sallesma: I have no memory of this 😱

I can bring it back and add it to the home page

guillaumemoby: Adélie is right. We want to keep the training list below the predefined program

sallesma: It's back in the game

guillaumemoby: I am not 100% sure about this but should we want to keep it the new button you have added on the homage into the 'training programs' page ?
pagetrainingmartin

Here is what I've added in mind

src/Navigation/Navigation.js
```diff
@@ -29,6 +30,7 @@ export const Navigation = () => (
       options={({ route }) => ({ title: route.params.drill.title })}
     />
     <Stack.Screen name="TrainingListPage" component={TrainingListPage} options={{ title: 'Training sessions' }} />
+    <Stack.Screen name="ProgramsPage" component={ProgramsPage} options={{ title: 'Choose a program for your team' }} />
```
Adeloche, line 33: On my phone, the title is cut ('Choose a program for your te')

src/Components/ProgramsPage.js
Adeloche: I agree with Moby, I think it is quite confusing to have both 'Trainings' and 'Training programs' on homepage (though maybe not so import for a V0). What do you think Martin ?

src/Components/HomePage.js
```diff
@@ -49,7 +57,7 @@ const styles = StyleSheet.create({
     backgroundColor: theme.BACKGROUND_COLOR_LIGHT,
   },
   menuItem: {
-    height: '30%',
+    height: '20%',
```
Adeloche, line 60: 25% maybe ? i have a quite weird white banner at the bottom of my homepage now

src/Components/ProgramsPage.js
sallesma: Yep, I'll move it to the programs page as Moby suggested above

src/Navigation/Navigation.js
sallesma: Then I can put only "Choose a program"

src/Components/ProgramsPage.js
sallesma: I just tweaked it a bit because it did'nt look like a cta

Looks like this (probably not perfect but i think it's ok enough for now?)

Screenshot 2020-05-02 at 21 14 05

src/Components/HomePage.js
sallesma: I weent back to 30% since the home page does not change in the end :)

PR #82, opened 2020-04-28
Fix warnings in jest tests

aenario:

There is a bunch of warnings in the jest console. [examples](https://github.com/UltimateApp/UltimateApp/runs/624061827?check_suite_focus=true#step:5:11)

 this PR removes them.

### DIFF
--- a/src/Components/DrillPage.js
+++ b/src/Components/DrillPage.js
@@ -1,4 +1,4 @@
-import React, { useRef } from 'react';
+import React, { useRef, useLayoutEffect } from 'react';
 import {
   ScrollView,
   StyleSheet,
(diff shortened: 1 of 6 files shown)


src/Components/DrillPage.js
```diff
@@ -59,9 +59,11 @@ export const DrillPage = props => {
     );
   };
 
-  navigation.setOptions({
-    headerRight: () => displayFavoriteButton(),
-  });
+  useLayoutEffect(() =>
```
sallesma, line 62: Why not using useEffect directly?

aenario: stupid copy paste from https://reactnavigation.org/docs/navigation-prop/#setoptions

sallesma: Should we swith to useEffect then or keep it this way?

aenario: From what I understand, useEffect is run asynchronously, potentially a few nanoseconds after the render, so it could cause a flash of non-modified header between the first render and the application of the effect on very slow device. whereas useLayoutEffect is run synchronously, ie not leaving time for a bad render to reach the screen.

So let's keep it this way.

PR #88, opened 2020-05-02
Training overview

sallesma:

Screenshot 2020-05-02 at 16 37 05

I haven't implemented the swipe gesture, not sure if it's important


src/Components/TrainingPage.js
```diff
@@ -18,28 +18,59 @@ export function getTrainingMinimalPlayersNumber(training) {
...
+      <View style={styles.programNavigation}>
+        {isFirstTraining && (
+          <TouchableOpacity style={styles.btnPrevNext} onPress={onPrevPress}>
+            <Text style={styles.btnPrevNextContent}>{'<'}</Text>
+          </TouchableOpacity>
+        )}
+        <Text style={styles.title}>Training session {currentTrainingIndex + 1}</Text>
+        {isLastTraining && (
+          <TouchableOpacity style={styles.btnPrevNext} onPress={onNextPress}>
+            <Text style={styles.btnPrevNextContent}>{'>'}</Text>
```
sallesma, line 45: I think it should be nicer to use an icon, or a bigger font size for `<` and `>`

guillaumemoby: Could it be possible to delete the words "minutes" and "players" or to keep it in the line with "12+" and "50+"?

guillaumemoby: Having the training name to replace "training session 1" and the program name to replace "coed 2 long" would be better in my opinion.

sallesma: I changed it as you say, but I kep the training title in the header when browsing trainings (not from a program)

From a program :
Screenshot 2020-05-03 at 13 52 37


From the trainings list :
Screenshot 2020-05-03 at 13 50 02


We loose the idea of progression within a program. You an navigate the trainings but you loose track of where you are within the program

PR #93, opened 2020-05-05
Add bravo page

sallesma:

Adding a bravo page at the end of a training
Not beautiful but I just wanted to add it before tonight's meeting because it is one of the last missing views from the wireframe

We'll make it better when the UI is ready

Screenshot 2020-05-05 at 14 00 07


src/Components/TrainingBravoPage.js
```diff
@@ -0,0 +1,84 @@
...
+function getGoals(training) {
+  return Array.from(new Set(training.drills.reduce((acc, { goals }) => acc.concat(goals), [])));
+}
+
+export const TrainingBravoPage = props => {
+  const { navigation, route } = props;
+  const { training, program } = route.params;
+
+  const finishTraining = () => {
+    props.completeTraining({ training, program });
```
sallesma, line 18: This could also be called after rendering ? 🤔

aenario: I think it would be better, in case the coach close the app after getting to this page.

sallesma: It's done

PR #95, opened 2020-05-06
Cleaner presentation, removing the youtube link

guillaumemoby:


src/Components/DrillIllustration.js
```diff
@@ -84,6 +85,14 @@ const DrillIllustration = props => {
     );
   };
 
+  const displayLine = () => {
+    if (props.drill.steps.length <= 1) {
+      return <View />;
+    } else {
+      return <View style={styles.lines} />;
```
sallesma, line 92: I think you could do something like `{props.drill.steps.length === 0 && <View style={styles.lines} />}` instead of extracting a function juste to have the condition

src/Components/DrillPage.js
```diff
@@ -19,8 +19,8 @@ import { toggleFavorite } from '../Store/Actions/favoriteAction';
 
 import theme from '../styles/theme.style';
 
-import iconFavoriteEmpty from '../../assets/ic_favorite_border.png';
-import iconFavoriteFull from '../../assets/ic_favorite.png';
```
sallesma, line 23: Then maybe remove the images if we d'ont use them anymore?

src/Components/shared/ButtonLight.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import { StyleSheet, Text, TouchableOpacity } from 'react-native';
+
+import theme from '../../styles/theme.style';
+
+const ButtonLight = props => {
```
sallesma, line 6: If the only difference between Button and ButtonLight is the color, we should have just one component and use a prop to pass like `light={true}` to choose when to have the light version

src/Components/DrillIllustration.js
guillaumemoby: I will discuss about this with Adélie. Thanks

src/Components/shared/Button.js
```diff
@@ -15,17 +18,38 @@ const styles = StyleSheet.create({
...
+    justifyContent: 'center',
+    alignItems: 'center',
   },
   text: {
     textAlign: 'center',
     color: theme.COLOR_PRIMARY_LIGHT,
     fontSize: theme.FONT_SIZE_MEDIUM,
   },
+  buttonLight: {
+    left: '10%',
```
aenario, line 35: Prop-tip, you can use array-of-styles to avoid repetition in the styles. 
This is not a problem for this component, as it is pretty simple, but styles often get big, so avoiding repetition is good.

```jsx
const Button = props => {
  const btnColor = props.buttonLight ? styles.buttonLight : styles.buttonDark;

  return (
    <TouchableOpacity style={[styles.button, btnColor]} onPress={props.onPress} testID="button">
      {/* ...*/}
    </TouchableOpacity>
  );
};

const styles = StyleSheet.create({
  button: {
    left: '10%',
    right: '10%',
    /*... everything that is common */
  },
  buttonLight: {
    backgroundColor: theme.COLOR_PRIMARY,
  },
  buttonDark: {
    backgroundColor: theme.COLOR_PRIMARY_LIGHT,
    borderColor: theme.COLOR_SECONDARY,
  }
  /*...*/
});
```

PR #97, opened 2020-05-07
Center video when there is only one step and unmute some video

guillaumemoby:


guillaumemoby:
I have a problem with the tests. Anyone can help ?

src/Components/VimeoVideo.js
```diff
@@ -62,7 +64,7 @@ const VimeoVideo = ({ vimeoId, screenWidth }) => {
   const renderError = () => {
     return (
       <View style={styles.spinnerStyle}>
-        <Text>Oopsie! There was an error loading the video...</Text>
+        <Text> Oopsie! There was an error loading the video...</Text>
```
aenario, line 67: You added a space here, you need to remove it or also add it in the test's `getText`

src/Components/DrillIllustration.js
```diff
@@ -168,6 +168,15 @@ const DrillIllustration = props => {
             </View>
           </>
         )}
+        {props.drill.steps.length <= 1 && (
+          <>
+            <View style={styles.videoAlone}>
```
sallesma, line 173: Have you tried to add `styles.videoAlone` to `<View style={{ height: 250 }}>`

If it worked it would be much cleaner than duplicated the markup :)

guillaumemoby: would be good to add the use "useNativeControls" to control the video for the long one but I haven't succeed to do it

sallesma: To prevent duplicating the call to the VimeoVideo component, you could now use it with a condition directly in the style of the view

And keep it above like it was before, with something like `<View style={[{ height: 250 }, props.drill.steps.length === 0 ? styles.standalone : null]}>` ?

```
            <View style={[{ height: 250 }, props.drill.steps.length === 0 ? styles.standalone : null]}>
              <VimeoVideo vimeoId={illustrationSource} screenWidth={screenDimension.width} sounds={sounds} />
            </View>
```

And even better, since we are using the same condition several times (at least 2), we could extract it and declare it in a variable juste before the `return`

```
const isStandalone = props.drill.steps.length === 0;
return (
  ...
  <View style={[{ height: 250 }, isStandalone ? styles.standalone : null]}>
  ...
  {!isStandalone && (
  ...
):
```

guillaumemoby: Pourquoi props.drill.steps.length === 0 au lieu de props.drill.steps.length <= 1 ?
Je veux que quand il y a qu'un seul step que ma vidéo soit centrée

sallesma: Dans ce cas `props.drill.steps.length === 1` non ? (si tu veux tester qu'il n'y en a 1)

Mais c'est pas hyper important. Tu peux mettre <=1 si tu penses que c'est mieux
Mon point c'est surtout qu'il vaut mieux éviter de dupliquer l'appel au composant VimeoVideo si on peut :)

PR #98, opened 2020-05-07
Persist complete trainings and favorites

Adeloche:
We want to keep completedtrainings and favorites when user closes the app.


src/Store/configureStore.js
```diff
@@ -1,9 +1,29 @@
...
+};
+// Middleware: Redux Persist Persisted Reducer
+const persistedReducer = persistReducer(persistConfig, rootReducer);
+
+// Redux: Store
+const store = createStore(
+  persistedReducer,
   fixtures,
   window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__(),
+  applyMiddleware(createLogger()),
```
sallesma, line 23: Maybe we should add this only on development environment ?

App.js
```diff
@@ -37,13 +38,15 @@ const App = props => {
...
   return (
-    <Provider store={Store}>
-      <NavigationContainer
-        initialState={initialState}
-        onStateChange={state => AsyncStorage.setItem(PERSISTENCE_KEY, JSON.stringify(state))}
-      >
-        <Navigation />
-      </NavigationContainer>
+    <Provider store={store}>
+      <PersistGate loading={null} persistor={persistor}>
```
sallesma, line 42: Have you experienced any delay in app loading ?

I just read this on the readme :

> This delays the rendering of your app's UI until your persisted state has been retrieved and saved to redux

Should we provide a fallback? Or maybe it's a problem for later?

src/Store/configureStore.js
Adeloche: Ok I tried, using composeWithDevTools with does this if I understood the doc correctly.

App.js
Adeloche: I haven't noticed any delay for now, though I guess it could happen when we will get more programs.
I don't fully understand the pros and limitations of this component, @aenario do you know what are the good practices when retrieving data ? 
- About delay,I could add a loading animation, or maybe you know a better way to achieve this ?
- Also I wonder if storing each {program,training} pairs with full objects in state completeTrainings is necessary and if it could be faster storing only id (something like {programId: [completedTrainingsIds]}).
- I'm not sure on the best way to handle the fact that, for example, a training content may have changed since it was stored as coplemtedTraining: so I still want it in state.completedTraining  but the content of the training won't match
What do you guys think ?

aenario: This will only delay until the asyncstorage is ready, which we already do anyway for the navigation thingy. 

We can optimize to do both in parallel and handle error but I think it can be done later.

To store only the IDs, we will need to use a persist-redux transformer https://github.com/rt2zz/redux-persist/blob/master/README.md#transforms
But this is also an optimization which can be done for v1

src/Store/configureStore.js
aenario: As it is now, this line does nothing I think. `createStore` only takes 3 parameters.

Adeloche: yes, forgot to push :)

App.js
Adeloche: Ok then I will try the transform thing in another PR.
Maybe for later version (after V1) it will be better to store this in database anyway, so that user can get his stored information from any device

PR #100, opened 2020-05-07
Program page

sallesma:

Screenshot 2020-05-07 at 22 01 49

Screenshot 2020-05-07 at 22 01 56


src/Components/ProgramPage.test.js
```diff
@@ -0,0 +1,28 @@
...
+    const tree = renderer
+      .create(
+        <Provider store={store}>
+          <ConnectedProgramPage route={route} />
+        </Provider>,
+      )
+      .toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+});
```
Adeloche, line 28: Ajouter un test quand il y a des completeTraining dans le program ? Et peut etre un test de navigation ?

src/Components/programs/Program.js
```diff
@@ -15,23 +18,29 @@ export const Program = props => {
   const onPress = () => {
     const firstTodoTraining = trainings.find(training => !programCompleteTrainings.includes(training));
     if (firstTodoTraining !== undefined) {
```
Adeloche, line 20: Do we still need that if and the fistToDoTraining computation ?

src/Components/programs/Program.js
```diff
@@ -15,23 +18,29 @@ export const Program = props => {
...
-      <View style={styles.progressBar}>
-        <View style={[StyleSheet.absoluteFill, styles.fillProgressBar, { width }]} />
-      </View>
-    </TouchableOpacity>
+    <ListItem>
+      <TouchableOpacity style={styles.program} onPress={onPress}>
+        <View style={styles.presentationContainer}>
+          <Text style={styles.programTitle}>{title}</Text>
+          <Text style={styles.completion}>
+            {completeTrainingsCount}/{trainings.length} trainings
```
Adeloche, line 32: Its detail and won't happen because programs will always have several trainings I guess, but I don't like this training**s** when there is only one program

sallesma: I guess we could include pluralization management, but maybe later ? (june version)

src/Components/programs/Program.js
sallesma: yep you're right

src/Components/ProgramPage.test.js
sallesma: Yep

PR #103, opened 2020-05-08
Make animation color like mockup

aenario:

Screenshot_20200508-171720


src/Components/Animation.js
```diff
@@ -332,7 +332,7 @@ class Animation extends React.Component {
 
     return (
       <View style={[styles.mainContainer, { height: this.animationHeight }, { width: this.animationWidth }]}>
-        {this._display(this.cuts)}
+        {!this.props.readonly && this._display(this.cuts)}
```
sallesma, line 335: Not sure we should hide the cuts in the app. I think it helps understand the movements of the players

guillaumemoby: As on ultiplays I am on the deleting the cut

sallesma: Ok, not so important to me, let's go this way

sallesma: (Just for the record and later discution, the cuts are displayed on ultiplay at the end)

Screenshot 2020-05-08 at 22 19 43

PR #105, opened 2020-05-09
More Drills and better programs

guillaumemoby:


src/Fixtures/index.js
```diff
@@ -1410,33 +1705,112 @@ const firstWeekFitnessTraining = {
...
   image: 'https://zupimages.net/up/20/19/su0s.jpg',
   description:
     "This training session aims at practicing your mark skills. Don't be broken to easily, generate dangerous passes",
-  drills: [warmupGameDrill, chevronDrill, markDrill, breakDrill, game3v3],
+  drills: [warmupGameDrill, dwarfDuelDrill, kyeDrill, breakDrill, normalGame],
 };
 
 const coed2LongTraining = {
   id: 4,
   illustrationType: 'Martin',
```
sallesma, line 1717: Do we need this attribute on trainings? I think we could remove it

Or maybe call it `author`, it looks like a mistake

PR #106, opened 2020-05-09
Add Favorites drills filter 

Adeloche:
Give user some way to select favorite drills in filters.

image


src/Components/FitnessFilters.js
```diff
@@ -10,15 +11,16 @@ import HeaderButton from './shared/HeaderButton';
...
     this.state = {
+      selectedFavorites: false,
       selectedLevels: [],
       selectedIntensities: [],
       selectedEquipmentLabels: [],
       selectedSeasonTimings: [],
       selectedGoals: [],
       durationInMinutes: undefined,
       displayedDrills: this.props.route.params.initialData,
+      favoriteDrills: this.props.favoriteDrills,
```
aenario, line 23: no need to put something in state if the composant does not change it.

src/Components/FitnessFilters.js
```diff
@@ -50,6 +61,8 @@ class FitnessFilters extends React.Component {
     } = this.state;
     let newData = this.props.route.params.initialData;
 
+    if (selectedFavorites)
+      newData = newData.filter(drill => favoriteDrills.map(favorite => favorite.id).includes(drill.id));
```
sallesma, line 65: I think we parse all the favorite `favoriteDrills` array for each drill in `newData` maybe we could do it once before the filter instruction ? (since it is user input, it can become quite large)

src/Components/FrisbeeFilters.js
```diff
@@ -37,9 +47,12 @@ class FrisbeeFilters extends React.Component {
   }
 
   applyFilters() {
-    const { selectedLevels, selectedGoals, numberOfPlayers } = this.state;
+    const { favoriteDrills } = this.props;
+    const { selectedFavorites, selectedLevels, selectedGoals, numberOfPlayers } = this.state;
     let newData = this.props.route.params.initialData;
 
+    if (selectedFavorites)
+      newData = newData.filter(drill => favoriteDrills.map(favorite => favorite.id).includes(drill.id));
```
sallesma, line 55: Same here

src/Components/FrisbeeFilters.test.js
```diff
@@ -36,26 +39,33 @@ describe('<FrisbeeFilters />', () => {
       const DummyScreen = props => null;
       const Stack = createStackNavigator();
 
+      const mockStore = configureMockStore()({
```
sallesma, line 42: I did something different in DrillPage.test.js (I guess I had the same problem)

If it is cleaner to use this package, maybe we could use it everywhere?

src/Components/FitnessFilters.js
```diff
@@ -93,6 +107,9 @@ class FitnessFilters extends React.Component {
       <View style={filterStyle.wrapper}>
         <Text style={filterStyle.counter}>{this.state.displayedDrills.length} drills available</Text>
         <ScrollView contentContainerStyle={filterStyle.filters}>
+          <View style={filterStyle.filter}>
+            <Button title="Favorite Drills" onPress={() => this.onFavoritesChange()} active={selectedFavorites} />
```
sallesma, line 111: The rendering looks a bit like a button to open something

Maybe we could use a checkbox ?
And a wording more like "My Favorites only"?

src/Components/FitnessFilters.js
Adeloche: j'ai essayé qqchose, mais je ne suis pas sure que c'est ce que tu avais en tete.

src/Components/FitnessFilters.js
aenario: Done with wording change, we can change to checkbox later on depending on feedback

src/Components/FitnessFilters.js
aenario: Je l'ai fait dans le 68d8feaa8109b6d41bfc34bfae064bb4f6f7ab2a

Issue #115, opened 2020-05-10
MinimalDrillPage title and timeline covers back button


sallesma:
### Environment

iOS

### Steps to Reproduce

Open the app on iOS
Go to any training
Start the training

### Expected Behavior

The back button and the drill titile do not overlap

### Actual Behavior

<img width="464" alt="Screenshot 2020-05-10 at 10 41 38" src="https://user-images.githubusercontent.com/3810795/81494656-d989be00-92aa-11ea-8dfb-ace2f5247569.png">

sallesma:
It was fixed by Romain with #117 

<img width="505" alt="Screenshot 2020-05-22 at 23 26 07" src="https://user-images.githubusercontent.com/3810795/82710655-a91b2a00-9c83-11ea-84f4-60c94d23c4e5.png">

PR #117, opened 2020-05-10
Fix header superposition on iOS

aenario:

Ref: #115 , not using auto-fix commit because I can't try on iOS

### DIFF
--- a/src/Components/DrillPageMinimal.js
+++ b/src/Components/DrillPageMinimal.js
@@ -41,7 +41,6 @@ export const DrillPageMinimal = props => {
     );
     navigation.setOptions({
       headerTitleAlign: 'center',
-      headerBackTitleVisible: false, // hide the "Back" text on iOS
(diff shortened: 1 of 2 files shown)


src/Components/DrillPageMinimal.js
```diff
@@ -41,6 +41,7 @@ export const DrillPageMinimal = props => {
     );
     navigation.setOptions({
       headerTitleAlign: 'center',
+      headerBackTitleVisible: false, // hide the "Back" text on iOS
```
sallesma, line 44: I think I had it on one or two more pages

I guess we should also hide it on every page where the title is dynamic (drill/training/program title ?)

sallesma: Or maybe everywhere to write it just once on the Navigation component

aenario: Done everywhere. We can tweak it on V1.

PR #122, opened 2020-05-13
Accordion for the program page

guillaumemoby:


guillaumemoby:
I have a problem of having a flatlist inside a scrollview. If anyone can have a look

src/Components/programs/Program.js
```diff
@@ -1,41 +1,39 @@
...
-    props.navigation.navigate('ProgramPage', { program: props.program });
-  };
-
   const width = `${(completeTrainingsCount * 100) / trainings.length}%`;
   return (
-    <ListItem>
-      <TouchableOpacity style={styles.program} onPress={onPress}>
-        <View style={styles.presentationContainer}>
+    <View style={styles.program}>
+      <ImageBackground source={{ uri: image }} style={styles.image} imageStyle={styles.imageOpacity}>
```
sallesma, line 18: I'm not a big fan of the background image on the programs. It feels overloaded, the screen is full of information

I would prefer to keep the minimalist UI we have right now (which I admit is not perfect)

src/Components/ProgramListPage.js
```diff
@@ -1,22 +1,101 @@
...
+          </View>
+          <View style={styles.containerAccordion}>
+            <Image style={styles.todoState} source={isDone ? doneImage : toDoImage} />
+          </View>
+        </TouchableOpacity>
+      </View>
+    );
+  };
+
+  const renderContent = (section, _, isActive) => {
```
sallesma, line 61: It would probably make it a lot more readable to extract most of this `renderContent` and `renderTraining` to another component (just as we did for `Program` last time)
Then it would be easier to understand

src/Components/programs/Program.js
guillaumemoby: I am ok with your choice. I will remove the imagebackground. I still have an issue with the flatlist inside the scrollview. I will be looking for a solution today

src/Components/ProgramListPage.js
```diff
@@ -1,23 +1,94 @@
...
+import arrowDark from '../../assets/arrow_dark.png';
 
 export const ProgramListPage = props => {
-  const { programs, navigation } = props;
+  const { navigation, programs } = props;
+
+  const [activeSections, setActiveSections] = useState([undefined]);
+
+  const setSections = sections => {
+    setActiveSections(sections.includes(undefined) ? [] : sections);
```
sallesma, line 20: Does the `sections` variable equals `[undefined]` at this time or is it just `[]`?

If it's `[]` we could directly call `    setActiveSections(sections);`

PR #123, opened 2020-05-15
Carousel animation for frisbee drill

guillaumemoby:


src/Components/DrillIllustration.js
```diff
@@ -260,14 +203,132 @@ const DrillIllustration = props => {
...
+        case IllustrationType.VIMEO:
+          return displayVimeo(props.drill.steps[activeIndex]);
+        default:
+          return <Text>No visual content for this drill</Text>;
+      }
+    }
+  };
+
+  const displayAnimation = ({ illustrationSource, title }, index) => {
+    const isFirstStep = isMultipleSteps ? index === 0 : true;
```
sallesma, line 226: I guess you don't need to check `isMultipleSteps` because `index === 0` is always the condition for being the first step?

src/Components/DrillIllustration.js
```diff
@@ -260,14 +203,132 @@ const DrillIllustration = props => {
...
+    const isLastStep = isMultipleSteps ? index === props.drill.steps.length - 1 : true;
```
sallesma, line 227: Same here

src/Components/DrillIllustration.js
```diff
@@ -120,49 +106,6 @@ const DrillIllustration = props => {
...
-                      <Image style={styles.buttonNext} source={buttonValidationGradient} />
-                    </Animated.View>
-                  </TouchableOpacity>
-                </View>
-              </View>
-            </View>
-            <View style={styles.lines} />
-          </>
-        )}
-        <Text style={styles.instruction}>{instruction}</Text>
```
sallesma, line 161: Maybe we should keep this one as long as we d'ont have text linked to animation frames

src/Components/DrillIllustration.js
```diff
@@ -260,14 +203,132 @@ const DrillIllustration = props => {
...
+          backgroundColor: theme.COLOR_SECONDARY,
+        }}
+        inactiveDotOpacity={0.4}
+        inactiveDotScale={0.6}
+      />
+    );
+  };
+
+  const drillTypes = () => {
+    if (props.drill.type === 'frisbee') {
```
sallesma, line 305: Use the enum here `props.drill.type === DrillTypes.FRISBEE`

src/Components/DrillIllustration.js
```diff
@@ -260,14 +203,132 @@ const DrillIllustration = props => {
...
+            renderItem={renderStep}
+            onSnapToItem={index => setActiveIndex(index)}
+          />
+          <View style={styles.pagination}>{pagination()}</View>
+        </SafeAreaView>
+      );
+    } else {
+      return (
+        <GestureRecognizer style={styles.container} onSwipeLeft={checkAnimationFast} config={swipeConfig}>
+          {checkSwitch()}
```
sallesma, line 323: This if/else statement means that frisbee drills are ALWAYS going to have animation (displayed within a carousel)
And fitness drils may have different illustrations : animation (not in a carousel), youtube and vimeo

Are we sure about this?
If yes, I think we should split this component and have 1 component for FrisbeeDrillIllustration and another for FitnessDrillIllustration, because the display is now different. That would make it easier to understand what's happening

guillaumemoby: Yes I am not 100% sure about this. I think if in the future there is a drill explain with a video and a animation we should use something else than the carousel. The problem now is, if we are explaining a drill with 2 videos. they both start when we arrive on the drillpage.

guillaumemoby: After reflexion, a frisbee drill will always use a carousel. If there is video inside ,these videos have to on "pause" at the begining

sallesma: Yep, I agree, when we add controls on video, the user will be able to play them when he's ready

PR #124, opened 2020-05-18
Tab navigation Homepage

guillaumemoby:


assets/HomePage/example.ai
```diff
@@ -0,0 +1,2008 @@
+%PDF-1.5%����
```
sallesma, line 1: Are you sure you want to add this file to the repo ?

package.json
```diff
@@ -14,6 +14,8 @@
   },
   "dependencies": {
     "@react-native-community/masked-view": "0.1.6",
```
sallesma, line 16: Can you provide some info about all these packages? What does each of them bring us?

(The more packages we have, the heavier our app in the end. And it also brings more maintenance cost)

src/Components/HomePage.js
```diff
@@ -80,9 +176,36 @@ const styles = StyleSheet.create({
...
     width: '100%',
-    paddingBottom: 5,
+    flex: 1,
+    paddingRight: 10,
+  },
+  description: {
+    alignItems: 'center',
+    width: '70%',
+  },
+  teamDescription: {
```
sallesma, line 194: It looks like team related variables are not used anymore

src/Components/HomePage.js
```diff
@@ -2,53 +2,129 @@ import React from 'react';
...
+      </TouchableHighlight>
+    </View>
+  );
+};
+
+const Fitness = props => {
+  return (
+    <View style={styles.mainContainer}>
+      <TouchableHighlight
+        onPress={() => props.navigation.navigate('DrillListPage', { type: DrillTypes.FITNESS })}
```
sallesma, line 54: what are these going to be? Direct links to fitness programs? Or pre-filtered fitness drills list?

guillaumemoby: What I have in mind is direct link to fitness program (that we have to create)

guillaumemoby:
I would need some help for the tests

package.json
```diff
@@ -25,12 +26,15 @@
     "react-native": "https://github.com/expo/react-native/archive/sdk-37.0.0.tar.gz",
     "react-native-elements": "^1.2.7",
     "react-native-gesture-handler": "~1.6.0",
+    "react-native-ionicons": "^5.0.0-rc",
```
sallesma, line 29: It looks like this package is not used since you get icons from `'@expo/vector-icons'`, maybe remove it?

package.json
```diff
@@ -25,12 +26,15 @@
...
+    "react-native-paper": "^3.10.1",
```
sallesma, line 30: Same for `react-native-paper`

package.json
```diff
@@ -25,12 +26,15 @@
...
     "react-native-kill-packager": "^1.0.0",
     "react-native-reanimated": "~1.7.0",
     "react-native-safe-area-context": "0.7.3",
     "react-native-screens": "~2.2.0",
     "react-native-svg": "^12.1.0",
     "react-native-swipe-gestures": "^1.0.5",
+    "react-native-vector-icons": "^6.6.0",
```
sallesma, line 37: Do we need this package if we use `'@expo/vector-icons'` ?

src/Components/HomePage.js
```diff
@@ -2,53 +2,129 @@ import React from 'react';
...
+      initialRouteName="Frisbee"
+      activeColor={theme.COLOR_PRIMARY}
+      inactiveColor={theme.COLOR_SECONDARY}
+      barStyle={{ backgroundColor: theme.BACKGROUND_COLOR }}
+    >
+      <Tab.Screen
+        name="Frisbee"
+        component={HomeScreen}
+        options={{
+          tabBarLabel: 'Home',
```
sallesma, line 114: Is it really `Home` and `Fitness`? Or `Frisbee` and `Fitness`?

guillaumemoby: It's Frisbee and Fitness But I hav'nt found a frisbee icon yet

package.json
guillaumemoby: I am using it to have the fitness icon. I will try to use only one icon package

package.json
guillaumemoby: We need it to use the bottom tab

package.json
guillaumemoby: It's removed

package.json
sallesma: That would be better. We should probably choose the one with the most icons included and the most users on github, so that we're sure to always find what we're looking for and the project won't stop

sallesma: I just had a look and ionicons has 1k users and 25 stars (= likes)
react-native-vector-icons has 155k users and 13k stars, and it includes far more icons (including ionicons)

So maybe we should keep react-native-vector-icons?

package.json
sallesma: Uh, ok.
But then the documentation says we must configure babel not to include everything on production

PR #153, opened 2020-05-23
Add video controls for vimeo

Adeloche:
First easy step to add control to vimeo video.
Current limitations:
- cannot rotate to landscape on Android when video is fullscreen (will need to use another component for this)
- I think the issue mentioned that when want to have a next/prev button to switch step..The need is not very clear to me. Is it only for fitness drills, if we want to go to next step when fullscreen is activated ? I'd like to discuss with someone, the need is not clear to me.


sallesma:
We had this discussion with Bart just before releasing V0.

Play/pause, fullscreen, duration and progress bar are awesome, but the prev/next buttons are misleading because there is no previous or next video reachable using them, although there are previous and next steps in the drill.

It looked like there was no way to add a callback on these buttons with the native controls, so we decided to leave it for later.

We thought using a wrapper (such as https://github.com/ihmpavel/expo-video-player) could fix the issue for us but we haven't explored it.

What do you think would be better?

Adeloche:
> We had this discussion with Bart just before releasing V0.
> 
> Play/pause, fullscreen, duration and progress bar are awesome, but the prev/next buttons are misleading because there is no previous or next video reachable using them, although there are previous and next steps in the drill.
> 
> It looked like there was no way to add a callback on these buttons with the native controls, so we decided to leave it for later.
> 
> We thought using a wrapper (such as https://github.com/ihmpavel/expo-video-player) could fix the issue for us but we haven't explored it.
> 
> What do you think would be better?

Sorry, I wasn't aware of the discussion. By prev/next button do you mean the fast forward arrows ? Then ok, I will try with the video player component.

Do we also want to be able to go to next step video from fullscreen ?

sallesma:
Ok thanks for explaining what these  buttons are on Slack :D

Tried this on a longer video (eg. conditionning) and it makes sense. My whole text above is probably to throw away.

Maybe we could just use a prop to autoplay (shouldPlay I guess?) only if we're not on the first step, to prevent the videos starting before the user has scrolled down on the drill page?

guillaumemoby:
I feel we should split the videos in 2 categories.
- One is the small video (less than 15 secondes) where we don't need to add a control. These videos are mute and it would be awesome if we could have it full screen
- Second is the long video where we need to have the control. We need to have the sounds of the videos and these videos has to appears in full screen if need. To play these videos the user has to press a button.

PR #154, opened 2020-05-25
Animation3

ZacharieALES:
I fixed the player number.

I splitted DrillCuts.js in two:
- DrillCuts.js which contains the cuts data ;
- DisplayedCut.js which is the component which enables to represent the cuts.

I reduced the shift when droping an element by fixing the height ratio of the animation in the editor (there was two different values in the editor for that...).

I also cleaned a little how the elements are positioned in the editor (to remove hard coded values).

And I promise to make less changes for my next pull resquest ^^'.


src/Components/animation/AnimationEditor.js
```diff
@@ -50,14 +48,43 @@ class AnimationEditor extends React.Component {
   };
 
   onLayout = e => {
+    var cHeight = e.nativeEvent.layout.height;
```
sallesma, line 51: I don't know exactly what is the `c` for? Maybe rename the variable so that it is explicit ? (churrosHeight if `c` stands for churros)

src/Components/animation/AnimationEditor.js
```diff
@@ -50,14 +48,43 @@ class AnimationEditor extends React.Component {
...
+      width: cWidth,
+      height: cHeight,
+      dTop: e.nativeEvent.layout.y,
       dLeft: e.nativeEvent.layout.x,
     });
 
+    // Create the elements in the horizontal bar
+    this.initialElements = [];
+
+    var aWidth = cWidth * this.wRatio;
```
sallesma, line 65: Same here with a (I guess it is for animation?)

src/Components/animation/AnimationEditor.js
```diff
@@ -50,14 +48,43 @@ class AnimationEditor extends React.Component {
...
+    var aHeight = cHeight * this.hRatio;
+
+    debug('onLayout: aWidth/aHeight: ' + aWidth + '/' + aHeight);
+
+    var playerRadius = Math.min(aWidth, aHeight) / 12;
+    this.ddeTop = aHeight + 2.5 * playerRadius;
```
sallesma, line 71: dde only stands for https://fr.wikipedia.org/wiki/Direction_d%C3%A9partementale_de_l%27%C3%89quipement in my mind

Maybe use a more explicit variable name?

src/Components/animation/AnimationEditor.js
```diff
@@ -279,8 +295,7 @@ class AnimationEditor extends React.Component {
     });
   };
 
-  _createDE(deType, height, width) {
-    debug('create de with type: ' + deType);
+  _createDE(deType, playerRadius, top, left) {
```
sallesma, line 298: Same here

src/Components/animation/DisplayedCuts.js
```diff
@@ -0,0 +1,77 @@
...
+          : undefined}
+      </View>
+    );
+  };
+
+  render() {
+    return (
+      <Animated.View key="1" style={[{ position: 'absolute', left: 0, top: 0 }]} height="100%" width="100%">
+        {this.props.drillCuts.cuts.length >= this.props.step &&
+        this.props.drillCuts.cuts[this.props.step] !== undefined &&
```
sallesma, line 68: I think with the new version of js you can write `this.props.drillCuts.cuts[this.props.step]?.map(...)`

The `?.` will check that it is not null or undefined


You don't even need to check `this.props.drillCuts.cuts.length >= this.props.step` because if the step is too high, someArray[99999] returns undefined so it won't do anything

src/Components/animation/AnimationEditor.js
ZacharieALES: c was for current but not very explicit indeed.

src/Components/animation/AnimationEditor.js
ZacharieALES: Yes it is. I replaced by animation.

src/Components/animation/AnimationEditor.js
ZacharieALES: DDE was for displayed draggable element ^^'. I replaced it.

src/Components/animation/DisplayedCuts.js
ZacharieALES: Ah this is great!

Do I need to make a new pull request with my new commit?

PR #155, opened 2020-05-30
Theory page

guillaumemoby:


src/Navigation/Navigation.js
```diff
@@ -59,5 +60,12 @@ export const Navigation = () => (
         title: 'New Drill',
       })}
     />
+    <Stack.Screen
+      name="TheoryPage"
+      component={TheoryPage}
+      options={() => ({
+        title: 'Theory',
```
sallesma, line 67: We should use I18n here

src/Components/HomePage.js
```diff
@@ -46,6 +46,7 @@ export default class HomePage extends React.Component {
             text={I18n.t('homePage.editor')}
             buttonLight="true"
           />
+          <Button onPress={() => this.props.navigation.navigate('TheoryPage')} text="Theory" buttonLight="true" />
```
sallesma, line 49: It needs to use the new version of the homepage :)

src/Components/TheoryPage.js
```diff
@@ -0,0 +1,989 @@
...
+  const [modalVisible, setModalVisible] = useState(false);
+  const [selectedItem, setSelectedItem] = useState({
+    item: {
+      id: 0,
+      text: '',
+      definition: '',
+    },
+  });
+
+  const sections = [
```
sallesma, line 23: I feel like this should be extracted in another file, probably within the fixtures folder, and using the same structure as the drills so that we can then translate everything

src/Components/TheoryPage.js
```diff
@@ -0,0 +1,989 @@
...
+};
+
+export default function TheoryPage() {
+  return (
+    <Tab.Navigator
+      initialRouteName="Dictionary"
+      activeColor={theme.COLOR_PRIMARY}
+      inactiveColor={theme.COLOR_SECONDARY}
+      barStyle={{ backgroundColor: theme.BACKGROUND_COLOR }}
+      shifting
```
sallesma, line 776: It makes it a bit more unclear, since the user does not understand what the tabs are for without clicking them

src/Components/TheoryPage.js
```diff
@@ -0,0 +1,989 @@
...
+          <View>{displayTheory()}</View>
+        </View>
+      </View>
+    </View>
+  );
+};
+
+export default function TheoryPage() {
+  return (
+    <Tab.Navigator
```
sallesma, line 771: I'm afraid another bottom tab navigator could be misleading for the user. We need to test this code updated with the master branch to figure it out

Don't you think we could add a "Theory" tab on the home page, with 3 images as we did for fitness and frisbee?
(We would probably need to rename Frisbee to Training)

src/Components/TheoryPage.js
```diff
@@ -0,0 +1,989 @@
...
+              {displayDictionary()}
+            </View>
+          </View>
+        </View>
+      </View>
+    </View>
+  );
+};
+
+const Essential = () => {
```
sallesma, line 383: You should add a `theory` subfolder and put each of these 3 components in a separate component there

src/Components/TheoryPage.js
guillaumemoby: yes we will keep the title. I want to keep the tab to navigate

src/Components/TheoryPage.js
guillaumemoby: I am going to have a try

src/Components/TacticsPage.js
```diff
@@ -0,0 +1,305 @@
...
+import { View, StyleSheet, Text, Dimensions, Modal, TouchableHighlight, FlatList } from 'react-native';
+import { MaterialCommunityIcons } from '@expo/vector-icons';
+
+import VimeoVideo from './VimeoVideo';
+
+import theme from '../styles/theme.style';
+
+const screenDimension = Dimensions.get('window');
+
+const TacticsPage = () => {
```
sallesma, line 11: I did not change anything to this page because it is the same as EssentialPage. I think we should tweak it at the end when we're ok with everything else

src/Components/HomePage.js
```diff
@@ -93,28 +96,86 @@ const Fitness = props => {
...
+  useLayoutEffect(() =>
+    navigation.setOptions({
+      headerRight: () => displayEditor(),
+    }),
+  );
   return (
     <Tab.Navigator
       initialRouteName="Frisbee"
-      activeColor={theme.COLOR_PRIMARY}
+      activeColor={theme.COLOR_PRIMARY_LIGHT}
```
sallesma, line 153: It changes a lot the UI. The whole application has a light background color

Maybe we should discuss this with vdd

src/Components/HomePage.js
```diff
@@ -8,21 +8,24 @@ import bodyweight from '../../assets/HomePage/bodyweight.jpg';
...
+import { MaterialCommunityIcons, Ionicons } from '@expo/vector-icons';
 
 const Tab = createMaterialBottomTabNavigator();
 
 const HomeScreen = props => {
   return (
     <View style={styles.mainContainer}>
       <TouchableHighlight onPress={() => props.navigation.navigate('AnimationEditorPage')} style={styles.menuItem}>
-        <ImageBackground source={simulator} style={styles.image}>
+        <ImageBackground source={ourPlays} style={styles.image}>
```
sallesma, line 26: Maybe this should be the third item, because it is less important than drills and programs?

src/Components/HomePage.js
guillaumemoby: Yes, we will. I felle that the tab isn't obvious enough with the light background

sallesma:
I'm going to fix the tests and I think we're good to go.

We just need to improve the content and translate it, but it can come later

PR #159, opened 2020-06-02
Delete editor elements

ZacharieALES:
In the editor, enables to drag elements outside of the animation area in order to remove them from the animation.

Also, if a cut or a counter-cut is dragged outside of the animation area, it is automatically moved inside at the the border.


sallesma:
What do you think about adding a trash icon and sending the items directly to it with drag/drop?

ZacharieALES:
> What do you think about adding a trash icon and sending the items directly to it with drag/drop?

I have the feeling that it may be a little easier to understand for a new user but less efficient in term of overall user experience (it requires to drag to a specific point rather than drag anywhere outside).

I think I prefer the current version (for the ease of use and the minimalistic interface) but if you are convinced it is better I can do it.

In that case we need to decide what would happen if an element is dragged outside of the animation area but not on the trash. It could be a problem if it goes behind the progress bar or the title bar. A solution could be to move it to the closest position inside the animation area (similarly to the cuts).

src/utils/i18n.js
```diff
@@ -5,7 +5,7 @@ import en from './locales/en';
 import fr from './locales/fr';
 
 I18n.defaultLocale = 'en';
-I18n.locale = Localization.locale;
+I18n.locale = 'en';
```
sallesma, line 8: Just a reminder so that we don't forget to revert this line before merging

src/Components/animation/AnimationEditor.js
```diff
@@ -342,14 +409,33 @@ class AnimationEditor extends React.Component {
...
+  };
+
   displayStepDescription = () => {
     // TODO
   };
 
-  render() {
-    debug('render AE');
-    debug('this.currentStep: ' + this.currentStep);
+  displayStepDescription = () => {
```
sallesma, line 434: This is a duplicate of the above method?

Maybe we can remove them since we have no description for now?

src/Components/animation/AnimationEditor.js
```diff
@@ -310,7 +374,10 @@ class AnimationEditor extends React.Component {
 
     if (deType === 'triangle') key = 602;
 
-    if (deType === 'disc') key = 603;
+    if (deType === 'disc') {
```
sallesma, line 377: `deType` is a bit confusing, could we rename it to something comre meaningful ?

For example `displayedElementType` ? (if it is what de means)

src/Components/animation/Drill.js
```diff
@@ -56,37 +61,143 @@ class Drill {
...
+
+    return isEqual;
+  }
+
   log() {
     if (this.positions !== undefined && this.positions !== null) {
       /* For each step */
       for (var stepId = 0; stepId < this.positions.length; stepId++) {
-        debug('step ' + stepId);
+        console.log('step ' + stepId);
```
sallesma, line 179: You should keep using debug because it prints just like `console.log` if you activate it on your computer (switch commented line in `debug.js`) and it does not print anything in the bundled application when we release, which is better for performance

src/Components/animation/ProgressBar.js
```diff
@@ -11,226 +16,235 @@ class ProgressBar extends React.Component {
...
 
-    for (var i = 0; i < this.props.stepCount; i++)
-      // The first step is always visible
-      if (i === 0) this.dotsOpacity.push(new Animated.Value(1));
-      else this.dotsOpacity.push(new Animated.Value(0));
+  render() {
+    return (
+      <View style={[{ position: 'absolute', left: 0, top: 0 }]} height="100%" width="100%">
+        {/* Gray dots */}
+        {this.props.readonly
```
sallesma, line 36: You can rewrite the condition `{!this.props.readlonly && this.state.state...}` it is less verbose

src/Components/animation/ProgressBar.js
```diff
@@ -11,226 +16,235 @@ class ProgressBar extends React.Component {
...
+        {!this.props.readonly ? (
+          <View style={[StyleSheet.absoluteFill]} height="100%" width="100%">
+            <TouchableOpacity style={[styles.dot, { right: 60 }]} onPress={() => this.props.onStepAdded()}>
+              <Image style={styles.controlIcn} source={iconAdd} />
+            </TouchableOpacity>
+            <TouchableOpacity style={[styles.dot, { right: 35 }]} onPress={() => this.props.onStepRemoved()}>
+              <Image style={styles.controlIcn} source={iconMinus} />
+            </TouchableOpacity>
+            <TouchableOpacity style={[styles.dot, { right: 10 }]}>
+              <Image style={styles.controlIcn} source={iconTrash} />
```
sallesma, line 139: I think we could use icons directly, so that we don't have to maintain images in our assets folder for this.

If needed we can do it together

src/Components/animation/Animation.js
```diff
@@ -1,32 +1,30 @@
...
+import { StyleSheet, Easing, Animated, Dimensions, View, Image } from 'react-native';
 
 import DisplayedElement from './DisplayedElement';
-import DrillCuts from './DrillCuts';
 import DisplayedCuts from './DisplayedCuts';
 import ProgressBar from './ProgressBar';
 import Drill from './Drill';
 
-import iconPlay from '../../../assets/play.png';
-import iconPrev from '../../../assets/prev.png';
```
sallesma, line 11: I guess we can remove the prev and next buttons images from the project

src/Components/animation/Animation.js
```diff
@@ -137,377 +69,265 @@ class Animation extends React.Component {
...
-      completeSequence.push(Animated.parallel(this._getStepAnimation(stepId, true)));
-
-    Animated.sequence(completeSequence).start();
-
-    if (this.props.onStepChange !== undefined && this.props.onStepChange !== null)
-      this.props.onStepChange(this._stepCount());
   }
 
   /** Get back to the previous step */
   _previousStep() {
```
sallesma, line 81: Maybe we can remove this method since there is no "previous step" button now

src/Components/animation/Animation.js
```diff
@@ -137,377 +69,265 @@ class Animation extends React.Component {
...
-
-    /* The first animation in the sequence enables to update the current step */
-    stepAnimation.push(
-      Animated.timing(this.currentStepAV, {
-        toValue: stepId,
-        duration: this.state.stepLength,
-        easing: Easing.linear,
-        key: 0,
-      }),
+  goToStep = stepId => {
```
sallesma, line 100: Do you think we really need this function, or we could remove it and use directly `this.playStep` where it's called?

src/Components/animation/Animation.js
```diff
@@ -137,377 +69,265 @@ class Animation extends React.Component {
...
   };
 
-  _display(item) {
-    if (item !== undefined && item !== null) return item.render();
-    else return <View />;
-  }
-
   render() {
-    this.cutsArray = [];
+    //<MaterialCommunityIcons name="chevron-double-left" color={theme.COLOR_PRIMARY} size={26} />
```
sallesma, line 196: Maybe remove this line if we don't use it?

src/Components/animation/Animation.js
```diff
@@ -137,377 +69,265 @@ class Animation extends React.Component {
...
-        {this.props.editable && this.cuts !== undefined && this.cuts !== null ? (
-          <DisplayedCuts step={this.state.currentStep} drillCuts={this.cuts} />
+        {this.props.editable && !this.state.animationPlaying ? (
+          <DisplayedCuts
+            step={this.state.currentStep}
+            positionPercentToPixel={this._positionPercentToPixel}
+            animation={this.props.animation}
+            onMoveEnd={this.props.onCutMove}
+          />
         ) : (
```
sallesma, line 207: You can remove the undefined if you write the condition this way :

```jsx
        {this.props.editable && !this.state.animationPlaying && (
          <DisplayedCuts
            step={this.state.currentStep}
            positionPercentToPixel={this._positionPercentToPixel}
            animation={this.props.animation}
            onMoveEnd={this.props.onCutMove}
          />
        )}
```

It's closer to the "react" way of writing conditions within render

src/Components/animation/Animation.js
```diff
@@ -137,377 +69,265 @@ class Animation extends React.Component {
...
+              animationWidth={this.animationWidth}
+              animationHeight={this.animationHeight}
+              animation={this.props.animation}
+              currentStepAV={this.currentStepAV}
+              positionPercentToPixel={this._positionPercentToPixel}
+            />
+          );
+        })}
+
+        {/* {this.props.editable ? (
```
sallesma, line 229: Maybe remove this code if we don't need it anymore?

src/Components/animation/Animation.js
```diff
@@ -137,377 +69,265 @@ class Animation extends React.Component {
...
-        () => {
-          /* Set all the elements to their initial positions */
-          this._setCurrentPositions();
-
-          if (this.props.editable) this._initializeCuts();
-          this.progressBar = this._createProgressBar();
-        },
-      );
-    }
+    //   if (
```
sallesma, line 270: Same here

src/Components/animation/AnimationEditor.js
ZacharieALES: Yes it is a duplicate. Ok we can remove them.

src/Components/animation/AnimationEditor.js
ZacharieALES: Done

src/Components/animation/Drill.js
ZacharieALES: Ok

src/Components/animation/ProgressBar.js
ZacharieALES: Do you have an example where an icon is used instead of an image?

src/Components/animation/ProgressBar.js
ZacharieALES: Ok (this syntax feels strange but I am all for consision)

src/Components/animation/Animation.js
ZacharieALES: Yes we can. Initially, goToStep was:
- playing a step if the step was after the current step; and 
- moving immediately to a step if the step was before the current step;
but it was confusing.

src/Components/animation/Animation.js
ZacharieALES: Yep

src/Components/animation/Animation.js
ZacharieALES: Done

src/Components/animation/Animation.js
ZacharieALES: Done

src/Components/animation/Animation.js
ZacharieALES: Done.

src/Components/animation/Animation.js
sallesma: ah ok

PR #160, opened 2020-06-02
English review with scott

sallesma:


sallesma:
We've done a bit more of half the drills, but that's already enough to be worth a PR

PR #168, opened 2020-06-17
New programs

guillaumemoby:


src/Components/drills/FrisbeeDrillIllustration.js
```diff
@@ -79,12 +80,25 @@ const FrisbeeDrillIllustration = props => {
...
+    }
+  };
+
+  const displayAnimation = ({ illustrationSource, instruction, videoIllustration }) => {
     return (
       <>
         <View style={styles.contentWrapper}>
           <Animation widthRatio={1} heightRatio={1 / 2} animation={new Drill(illustrationSource)} />
         </View>
+        <View>{displayVideo(videoIllustration)}</View>
```
sallesma, line 101: Why don't you use a step with a video to keep it within the carousel?

It seems to me that we are breaking here the paradigm of a drill being made of steps and moving against KISS principle (https://en.wikipedia.org/wiki/KISS_principle)

src/Fixtures/en.js
```diff
@@ -731,64 +766,6 @@ const breakDrill = {
   ],
 };
 
-const threeColumnsLongDrill = {
```
sallesma, line 734: Tou need to apply the same changes to the French version so that it stays up-to-date

Espacially if you're visiting someone from the french federation next week, he'll probably have his phone in french

src/Components/drills/FrisbeeDrillIllustration.js
guillaumemoby: The carroussel is slowing down everything and Here is the video to illustrate one step. It could look as another step if we have to swipe to see it.

src/Fixtures/en.js
guillaumemoby: Yes I Know, I didn't have the time yet. Could we merge the PR and I will take the time tomorrow to work on the french version

PR #174, opened 2020-06-22
data fitness

guillaumemoby:


sallesma:
The UX/UI is not perfect but we have our three fitness views. We now need some bodyweight programs, and to translate to french

PR #185, opened 2020-07-24
better splash and name

sallesma:

Must check if it breaks installation of something


sallesma:
The problem seems to come from the new splash screen. Maybe it too big or something...

PR #195, opened 2020-08-19
Refactor some animation components

sallesma:

- Switched MovingCircle and DraggableDisplayedElement to functional components
- Made DisplayedCuts simpler


src/Components/animation/DisplayedCuts.js
```diff
@@ -21,83 +21,58 @@ class DisplayedCuts extends React.Component {
...
-        {this.state.stateFromProps.cuts !== undefined && this.state.stateFromProps.cuts !== null
-          ? this.state.stateFromProps.cuts[this.state.step]?.map(item => {
-              return (
-                <Animated.View
-                  key={item.key}
-                  style={[{ position: 'absolute', left: 0, top: 0 }]}
-                  height="100%"
-                  width="100%"
-                >
-                  {this.state.stateFromProps.cuts[this.state.step]?.map(this._displayCut)}
```
sallesma, line 74: I don't know why we had this double nested loop.. It looks to me like we rendered n times each cut instead of once

src/Components/animation/DisplayedCuts.js
```diff
@@ -21,83 +21,58 @@ class DisplayedCuts extends React.Component {
...
+      <Animated.View style={{ position: 'absolute', left: 0, top: 0 }} height="100%" width="100%">
+        {this.state.stateFromProps?.cuts[this.state.step]?.map(this._displayCut)}
       </Animated.View>
     );
   }
 
   /** Used to update the cuts when a modification is made in the animation */
   static getDerivedStateFromProps(props, state) {
     // Test if the animation has changed
-    var isEqual = true;
```
sallesma, line 86: I used const and let instead of var because it makes it clear which ones are not updated (const) and the scope of let is simpler to get while reading the code

src/Components/animation/DisplayedCuts.js
```diff
@@ -193,36 +165,26 @@ const _initializeStateFromProps = props => {
...
-              var left2 = (x1 + x2 - d2) / 2;
-              var top2 = (y1 + y2) / 2;
+              const left1 = (x0 + x2 - d1) / 2;
+              const top1 = (y0 + y2) / 2;
+              const left2 = (x1 + x2 - d2) / 2;
+              const top2 = (y1 + y2) / 2;
 
               cuts[stepId].push({
                 key: elemId,
-                x0,
```
sallesma, line 203: Removed these unused params

Issue #197, opened 2020-08-22
Animation ProgressBar does not reset when the drill changes


sallesma:
- Open a fribess training (eg : vertical stack
- Play the animation of a drill
- Look at another drill using the top timeline
- the progressbar still matches the first animation

If the animation is shorter, extra steps appear :

<img width="427" alt="Screenshot 2020-08-22 at 12 08 16" src="https://user-images.githubusercontent.com/3810795/90953943-32ebb680-e470-11ea-8509-f596dc7dbbbe.png">

PR #203, opened 2020-09-23
Redux animation

ZacharieALES:
This PR add the possibility to locally save and manage animations made in the editor.


src/utils/locales/en.js
```diff
@@ -93,6 +93,28 @@ export default {
     tactics: 'Tactics',
   },
   animationEditor: {
+    untitledAnimation: 'Unamed animation',
```
sallesma, line 96: `Unnamed` ?

sallesma, line 96: And is it animation or drill ? (We should choose a word and stick to it, the home page button says "new drill")

src/utils/locales/fr.js
```diff
@@ -93,6 +93,29 @@ export default {
     tactics: 'Tactiques',
   },
   animationEditor: {
+    untitledAnimation: 'Animation sans titre',
+    drillManager: {
+      ok: 'OK',
+      overwriteAnimation: 'Ecraser une animation ?',
```
sallesma, line 99: É

src/utils/locales/fr.js
```diff
@@ -93,6 +93,29 @@ export default {
...
+      askOverwrite: 'Voulez-vous la remplacer  ?',
```
sallesma, line 100: double space

src/utils/locales/fr.js
```diff
@@ -93,6 +93,29 @@ export default {
...
+      save: 'Sauvegarder',
+      rename: 'Renommer',
+      saveLocally: 'Sauvegarder sur cet appareil',
+      load: 'Ouvrir une animation sauvegardée',
+      delete: 'Supprimer une animation sauvegardée',
+      cancel: 'Annuler',
+      clickHereToRename: 'Appuyer ici pour renommer',
+      cta: "Envoyer à l'équipe de développement",
+      enterTitle: "Entrer le titre de l'animation",
+      alreadyUsedTitle: 'Ce titre est déjà utilisé par une autre application',
```
sallesma, line 114: "une autre animation" ?

src/utils/locales/fr.js
```diff
@@ -93,6 +93,29 @@ export default {
     tactics: 'Tactiques',
   },
   animationEditor: {
+    untitledAnimation: 'Animation sans titre',
```
sallesma, line 96: "Animation" or "exercice" ?

src/Store/configureStore.js
```diff
@@ -8,7 +8,7 @@ import fixtures from '../Fixtures';
 const persistConfig = {
   key: 'root',
   storage: AsyncStorage,
-  whitelist: ['completeTrainings', 'favoriteDrills'],
+  whitelist: ['completeTrainings', 'favoriteDrills', 'customeAnimations'],
```
sallesma, line 11: customAnimation, without the 'e'

src/Store/Reducers/animationReducer.js
```diff
@@ -0,0 +1,34 @@
+const initialState = [];
+
+function animationReducer(state = initialState, action) {
+  let nextState;
+  switch (action.type) {
+    case 'SAVE_ANIMATION':
+      // If the saved animation previously had another title, remove the animation with this old title
+      let oldAnimationIndex = state.findIndex(item => item.title === action.value.oldTitle);
```
sallesma, line 8: It smells like this should be a "renameDrill" action since we do 2 different things here ?

src/Store/Reducers/animationReducer.js
```diff
@@ -0,0 +1,34 @@
...
+
+      if (oldAnimationIndex !== -1) nextState = state.filter((item, index) => index !== oldAnimationIndex);
+      else nextState = [...state];
+
+      // If the animation is already saved, replace it; otherwise add it
+      let animationIndex = state.findIndex(item => item.title === action.value.title);
+      if (animationIndex !== -1) {
+        nextState[animationIndex] = action.value;
```
sallesma, line 16: All these Array manipulations make me wonder if we should have an Array or an Object in the store. If we had an object with the form { drillName: drill }, all these accesses would be much more easier to do (and less complex)

I just don't know (but the internet probaably does) if it is a good idea to have both arrays and objects in our root reducer

src/Store/Reducers/animationReducer.js
```diff
@@ -0,0 +1,34 @@
...
+      } else nextState = [...nextState, action.value];
+
+      if (nextState !== undefined && nextState !== null) nextState.sort((a, b) => (a.title > b.title ? 1 : -1));
```
sallesma, line 19: Not sure we should sort it in the store. I think it's probably more a view's responsibility, so maybe we should do this when we display these drills

src/Components/animation/AnimationEditor.js
```diff
@@ -27,6 +26,7 @@ class AnimationEditor extends React.Component {
         triangle: 1,
       },
       isElementMoving: false,
+      currentStepAV: new Animated.Value(0), // Enables to update  the current step inside an animation
```
sallesma, line 29: double space

sallesma: Does this fix this bug? https://github.com/UltimateApp/UltimateApp/issues/197

src/Components/animation/AnimationEditor.js
```diff
@@ -309,6 +307,39 @@ class AnimationEditor extends React.Component {
       </View>
     );
   }
+
+  /** Used to update the animation when an animation is opened in AnimationEditorPage */
+  static getDerivedStateFromProps(props, state) {
```
sallesma, line 312: I'd like to think about a way not to use this function

src/Components/AnimationEditorPage.js
```diff
@@ -38,19 +100,328 @@ export const AnimationEditorPage = props => {
...
+    shadowOffset: {
+      width: 0,
+      height: 2,
+    },
+    shadowOpacity: 0.25,
+    shadowRadius: 3.84,
+    elevation: 5,
+  },
+  item: {
+    backgroundColor: '#d5f6ffff',
```
sallesma, line 399: It would be better to use colors from the theme

src/Components/AnimationEditorPage.js
```diff
@@ -38,19 +100,328 @@ export const AnimationEditorPage = props => {
...
+    padding: 10,
+    marginVertical: 8,
+    marginHorizontal: 10,
+  },
+  openButton: {
+    backgroundColor: '#F194FF',
```
sallesma, line 405: idem

src/Components/AnimationEditorPage.js
```diff
@@ -38,19 +100,328 @@ export const AnimationEditorPage = props => {
...
+              }}
+            >
+              <Text style={styles.textStyle}>{`${I18n.t('animationEditor.drillManager.cancel')}`}</Text>
+            </TouchableHighlight>
+          </View>
+        </View>
+      </Modal>
+
+      {/* Modal to delete an existing animations */}
+      <Modal visible={modalDeleteVisible}>
```
sallesma, line 250: I haven't read all the new code, but do you think we could  do the same with only one modal, and change it's content ?

src/Components/AnimationEditorPage.js
```diff
@@ -38,19 +100,328 @@ export const AnimationEditorPage = props => {
...
+    if (item.title === animationTitle) createNewAnimation();
+  };
+
   return (
     <View style={styles.animationEditorPage}>
-      <AnimationEditor onAnimationChange={saveAnimationState} />
+      <AnimationEditor onAnimationChange={saveAnimationState} animation={currentAnimationState} />
+
+      {/* Modal to manage the dirlls */}
+      <Modal visible={modalManagerVisible}>
```
sallesma, line 153: It feels like we should extract this to a new component, with an explicit name (such as CustomDrillManager or something better).

Maybe just leave the modal here

sallesma (commented):
Nice :)

I think I reviewed it in reverse order, so the most important comment are probably at the end

Also, we'll need to add some tests before merging

src/utils/locales/en.js
ZacharieALES: Ah oui tiens

ZacharieALES: I was wondering...  ok I will switch to drill.

src/Components/animation/AnimationEditor.js
ZacharieALES: Yes I know. I used it since I wanted to make it quick and I don't know how to do it another way.

src/Components/AnimationEditorPage.js
ZacharieALES: I guess we could. How do we do it? We have 1 component for each possible content and in the return() method of AnimationEditorPage we use the syntax {isMainModal? <DrillManagerModal/> : isTitleModal? <TitleDrillManagerModal/> : ...
 ?

sallesma: Or we could have a Modal which renders a component, and this component takes care of it all

Something like 
```jsx
      <Modal visible={isModalVisible}>
        <NewDrillManager animation={animation} blahblah={blehbleh}>
      </Modal>
```

So that the AnimationEditorPage stays small and easy to read

sallesma:
I am making a few changes

- [ ] Introduce a editor namespace to tidy up our components
- [ ] Make sure components and i18n keys are cleaned
- [ ] Better menus (discussed on slack)
- [ ] Better wording
- [ ] Try to have AnimationEditorPage not doing everything
- [ ] Add tests

guillaumemoby (approved):

PR #220, opened 2020-10-09
Attemps to fix offset for DraggableDisplayedElement

sallesma:

It would be nice to test it on an iPhone and several android devices

It works OK on my emulator


src/Components/animation/AnimationEditor.js
```diff
@@ -297,9 +287,8 @@ class AnimationEditor extends React.Component {
           onMoveStart={this.onMoveStart}
           onElementMoveEnd={this.onElementMoveEnd}
           onDimensionSet={this.setAnimationDimension}
-          onTopMarginSet={this.setAnimationTopMargin}
           onCutMove={this.cutMove}
-          widthRatio={1}
+          widthRatio={this.wRatio}
```
sallesma, line 291: It is equal to 1, but it's better to use the variable than a magic number

src/Components/animation/AnimationEditor.js
```diff
@@ -121,10 +114,7 @@ class AnimationEditor extends React.Component {
    */
   _positionPixelToPercent = (x, y) => {
     /* Here we assume that there is no view at the left or the bottom of the editor */
-    return [
-      (x - this.state.dLeft) / this.state.animationWidth,
-      (y - this.state.animationTopMargin) / this.state.animationHeight,
```
sallesma, line 126: It looks like it is taken into account in the setting of `dTop`, so we don't need it anymore

PR #225, opened 2020-10-14
Fix ios

guillaumemoby:


src/Components/animation/Animation.js
```diff
@@ -42,7 +46,11 @@ class Animation extends React.Component {
    * y2: corresponding vertical position in pixel (=0 if centered)
    */
   _positionPercentToPixel = (x, y) => {
-    return [this.animationWidth * x + this.dLeft, this.animationHeight * y + this.dTop];
+    if (Platform.OS === 'ios') {
```
sallesma, line 49: It looks like this is the same formula in either cases, since dleft  equals 0 on ios?

Maybe we could remove the if  to simplify this?

src/Components/animation/Animation.js
```diff
@@ -150,6 +154,10 @@ class Animation extends React.Component {
 
                 if (pageY !== undefined) this.dTop = pageY;
                 if (pageX !== undefined) this.dLeft = pageX;
+
+                if (Platform.OS === 'ios') {
+                  pageX = 0;
```
sallesma, line 159: It looks like it does not change anything here because we have already used pageX to modify this.dLeft

We should either do it before the assignation or remove it (if it's useless)

sallesma (commented):
There is this comment above you need to check.

And also check again on an iPhone because I removed some padding on the icons in the editor topbar and we must make sure the `-45` is still relevant

sallesma (approved):

PR #228, opened 2020-10-27
Extract elements

sallesma:

I extracted Player, Disc and Cone components to make sure they are consistent everywhere in the app


sallesma:
@guillaumemoby could you test it on an iphone?

guillaumemoby (approved):

PR #231, opened 2020-11-02
Share play

sallesma:

## Before releasing

We need to make sure the following procedure is done: https://docs.expo.io/versions/latest/sdk/document-picker/#managed-workflow
to have it work on ios


sallesma:
Ideas to discuss:
- save drill on import?
- save imported drills separately from own drills?
- how to handle sad path?
- add tests

PR #237, opened 2020-11-11
Numero instagram

guillaumemoby:

### DIFF
--- a/src/Components/drills/FrisbeeDrillIllustration.js
+++ b/src/Components/drills/FrisbeeDrillIllustration.js
@@ -18,11 +18,10 @@ export const IllustrationField = {
 
 const FrisbeeDrillIllustration = (props) => {
   const carouselRef = useRef(null);
-  const [activeIndex, setActiveIndex] = useState(0);
   const [illustrationPreference, setIllustrationPreference] = useState(IllustrationField.ANIMATION);
 
   useEffect(() => {
-    setActiveIndex(0);
+    carouselRef.current.snapToItem(0);
   }, [props.drill]);
 
   const renderTitle = (title, index) => {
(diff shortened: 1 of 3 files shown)


guillaumemoby:
We have a problem on ios but not on Android:
android: 
Screenshot_2020-11-16-09-21-43-256_host exp exponent
ios:

src/Components/drills/FrisbeeDrillIllustration.js
```diff
@@ -88,14 +89,26 @@ const FrisbeeDrillIllustration = (props) => {
...
+  };
+
+  const renderPagination = (index) => {
     return (
       <Pagination
         dotsLength={props.drill.steps.length}
-        activeDotIndex={activeIndex}
+        activeDotIndex={index}
         containerStyle={{
-          backgroundColor: theme.BACKGROUND_COLOR_LIGHT,
```
sallesma, line 97: I think your bug came from this line

It should be better with a transparent background

src/Components/drills/FrisbeeDrillIllustration.js
```diff
@@ -88,14 +89,26 @@ const FrisbeeDrillIllustration = (props) => {
...
+      );
+    }
+  };
+
+  const renderPagination = (index) => {
     return (
       <Pagination
         dotsLength={props.drill.steps.length}
-        activeDotIndex={activeIndex}
+        activeDotIndex={index}
```
sallesma, line 108: This breaks the effect used on line 21 : when you move to another drill in the top navigation, it should display the first step. If you use the index directly, it stays on the step it was displaying

guillaumemoby:
It's good on Iphone

sallesma (approved):

PR #245, opened 2020-11-11
better arrow

guillaumemoby:

### DIFF
--- a/src/Components/drills/FrisbeeDrillIllustration.js
+++ b/src/Components/drills/FrisbeeDrillIllustration.js
@@ -94,8 +94,8 @@ const FrisbeeDrillIllustration = (props) => {
     return (
       <>
         {renderTitle(item.title, index)}
+        <View style={styles.pagination}>{renderPagination()}</View>
         <ScrollView>
-          <View style={styles.pagination}>{renderPagination()}</View>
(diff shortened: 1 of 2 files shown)


src/Components/drills/FrisbeeDrillIllustration.js
```diff
@@ -94,8 +94,8 @@ const FrisbeeDrillIllustration = (props) => {
     return (
       <>
         {renderTitle(item.title, index)}
-        <View style={styles.pagination}>{renderPagination()}</View>
         <ScrollView>
+          <View style={styles.pagination}>{renderPagination()}</View>
```
sallesma, line 98: Are you sure about this change?

In minimaldrillpage,, if i scroll down to read the text, I don't see the dots anymore:

Capture du 2020-11-11 23-10-48

sallesma (commented):
Looks good to me

I just have this question about the dots position

src/Components/drills/FrisbeeDrillIllustration.js
guillaumemoby: no It was a mistake that I have to change. thks Martin

sallesma (approved):

Issue #238, opened 2020-11-11
Changing the minimal page


guillaumemoby:

guillaumemoby:
We should but the drill name with a counter in the title
Removing the i button
Adding the description of the drill at the bottom of the scrolldown

PR #249, opened 2020-11-11
Minimal

guillaumemoby:

Fixes https://github.com/disc-in/UltimateApp/issues/238

### DIFF
--- a/src/Components/DrillPageMinimal.js
+++ b/src/Components/DrillPageMinimal.js
@@ -1,7 +1,6 @@
 import React, { useLayoutEffect, useCallback } from 'react';
-import { Platform, StyleSheet, View, ScrollView, Text, TouchableOpacity } from 'react-native';
+import { Platform, StyleSheet, View, ScrollView, Text } from 'react-native';
 import { connect } from 'react-redux';
-import { MaterialCommunityIcons } from '@expo/vector-icons';
 
 import I18n from '../utils/i18n';
 import theme from '../styles/theme.style';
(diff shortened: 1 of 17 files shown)


src/Components/DrillPageMinimal.js
```diff
@@ -42,27 +43,35 @@ export const DrillPageMinimal = (props) => {
...
+          {I18n.t('drillPageMinimal.headerTitle', { trainingTitle: drill.title })}
         </Text>
-        <ProgressBar total={training.drills.length} current={currentDrillIndex + 1} onDotPress={onProgressDotPress} />
+        <View >
+          <ProgressBar total={training.drills.length} current={currentDrillIndex + 1} onDotPress={onProgressDotPress} />
+        </View>
       </View>
     );
+
+    const headerRight = () => {
```
sallesma, line 54: I think we decided (you wrote this in the issue #238 ) to remove this button and put the description at the bottom of the screen ?

src/Components/DrillPageMinimal.js
```diff
@@ -112,7 +121,7 @@ const styles = StyleSheet.create({
     marginLeft: '10%',
     marginTop: 15,
     marginBottom: 5,
-    width: '80%',
```
sallesma, line 115: I think the 80% was here to handle long titles

For example:

Capture du 2020-11-12 18-05-22


Some people have smaller phones than we do, and they might have the problem even if we don't have it

src/Components/DrillPageMinimal.js
guillaumemoby: I havn't succeed to to it. The other problem is the scrolldown on ios doesn't work well . We should change it

guillaumemoby:
It's working on Ios

sallesma (approved):

PR #260, opened 2020-11-17
Removing the Scrollview

guillaumemoby:


sallesma (commented):
Have you checked the 4 following cases on ios?

- frisbee drill on drillpage
- frisbee drill on drillminimalpage
- fitness drill on drillpage
- fitness drill on drillminimalpage

Because we also have nested scrolling on fitness drills and I'm wondering if we could have the same bug over there?

guillaumemoby:
It's good on all these 4 points on Ios 👍

sallesma (approved):

PR #276, opened 2020-12-18
Firebase sharing

sallesma:

## Todo

- [x] Fix the format of the imported drill (`[null]` seems   to disappear)
- [x] Manage secure accesses to firebase db
- [x] Build a nicer user interface
- [x] Open a play straight after importing it
- [x] Use new credentials since the one we use were committed on github
- [x] write tests for the importer
- [x] Fix timer warning
- [x] handle sad path
- [ ] ~~Do not import all firebase packages~~
- [x] Create issue to remember: shared plays shouldnt be overridden by recipients
- [x] Make sure env var are bundled into production build


package-lock.json
```diff
@@ -1450,6 +1430,348 @@
         "tiny-queue": "^0.2.1"
       }
     },
+    "@firebase/analytics": {
```
sallesma, line 1433: It would be good to add only the packages we need

src/utils/firebase.js
```diff
@@ -0,0 +1,21 @@
+import firebase from 'firebase/app';
+import 'firebase/database';
+
+//Initialize Firebase
+if (firebase.apps.length === 0) {
+  firebase.initializeApp(firebaseConfig);
```
sallesma, line 6: I did not commit the config object on this branch, we need to find a way to manage the credentials

src/utils/firebase.js
```diff
@@ -0,0 +1,21 @@
...
+}
+
+const reference = (uuid) => {
+  return firebase.database().ref(`customPlays/${uuid}`);
```
sallesma, line 10: Il y a ce warning
> Setting a timer for a long period of time, i.e. multiple minutes, is a performance and correctness issue on Android as it keeps the timer module awake, and timers can only be called when the app is in the foreground. See https://github.com/facebook/react-native/issues/12981 for more info.
(Saw setTimeout with duration 386136ms)

Plus d'explications sur la cause et les conséquences : https://github.com/facebook/react-native/issues/12981#issuecomment-652745831

Une solution possible serait d'utiliser l'api http directement : https://stackoverflow.com/questions/44603362/setting-a-timer-for-a-long-period-of-time-i-e-multiple-minutes/62446792#62446792
(mais c'est du taff)

Il y a ce workaround pour changer l'utilisation de setTimeout : https://github.com/firebase/firebase-js-sdk/issues/97#issuecomment-452568889

App.js
```diff
@@ -39,12 +41,23 @@ const App = (props) => {
...
+    },
+  };
+
   return (
     <ReduxProvider store={store}>
       <PersistGate loading={null} persistor={persistor}>
         <PaperProvider>
           <NavigationContainer
             initialState={initialState}
+            linking={linking}
```
sallesma, line 59: Le linking fonctionne correctement lorsque l'app est ouverte alors qu'elle était déjà en background.
En revanche ça ne fonctionne pas très bien lorsque l'app n'était pas déjà ouverte

- Issue semblable : https://github.com/react-navigation/react-navigation/issues/6738
- Tester avec un apk aussi

guillaumemoby (approved):

Issue #407, opened 2024-05-14
Videos not loading in app (Android 8)


JCapul:
### Environment

Android 8.0.0

### Steps to Reproduce

- Open the app
- Open any drill or exercise

### Expected Behavior

Videos loads and can be watched.

### Actual Behavior

The video container returns the following error message "Oups ! Une erreur est survenue pendant le chargement de la vidéo".

Diagrams works correctly though.

JCapul:
See snapshot below.

Note that the top status bar is also not shown, not sure it's related but it looks like a bug too.

![Screenshot_20240523-110908_Disc In](https://github.com/disc-in/UltimateApp/assets/37209478/eced95b4-33d3-450a-a274-48ed724f2ad8)

Thanks in advance for your help !

daveliepmann:
I'm getting the same video loading issues on iOS

sallesma:
Hi, thanks for the bug report. It looks like the vimeo videos are not loading anymore, probably dur to a change in the vimeo urls. Both moby and I have are quite busy at the moment, We'll try to fix it soon. In the meantime, feel free to have a look a the issuechecking out the code :)

sallesma:
It is probably caused by this line: https://github.com/disc-in/UltimateApp/blob/main/src/Components/shared/VimeoVideo.js#L16

c0c4i:
News on this? @JCapul @sallesma

hippalectryon-0:
Should be fixed in #418

PR #411, opened 2024-06-25
Fix VimeoVideo.js with updated Vimeo API

JCapul:
This is a quick and dirty attempt to implement the changes I could figured out from the revamped Vimeo API documentation (ref: https://developer.vimeo.com/api/files/video-links)

I am complete noob in mobile dev so I would need to spare some time to setup and understand the dev environment required to properly test and validate this change. So I just thought, it might be quicker to just push this PR, initiate the discussion, and maybe you will be able to check it much quicker than myself. 

### DIFF
--- a/src/Components/shared/VimeoVideo.js
+++ b/src/Components/shared/VimeoVideo.js
@@ -13,7 +13,7 @@ const VimeoVideo = ({ vimeoId, sounds, shouldPlay }) => {
   const [error, setError] = useState();
 
   useEffect(() => {
-    const vimeoUrlSource = `https://player.vimeo.com/video/${vimeoId}/config`;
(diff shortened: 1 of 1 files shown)


JCapul:
attempt to fix #407

JCapul:
Ok, at least I managed to run the tests and they all pass. Now, I just need to check if the fix actually work...sorry that's not really the proper way to do all this !

JCapul:
Ok, I finally managed to make Expo Go work and start digging into the problem. 
It looks like Vimeo now requires to have the app registered and to use an access token for the request. With the changes I have made to reflect the new API spec, I get an error 401 Unauthorized.

